PR #2, opened 2016-03-16
Prepend base url to request path

alexwdaniel:
It'd be nice not to have to write out the full url on each `visit`. In test_helper.exs:

```
{:ok, _} = Application.ensure_all_started(:wallaby)
Application.put_env(:wallaby, :base_url, Allocations.Endpoint.url)
```

then `visit("/sign_in")` would navigate to http://localhost:4001/sign_in.


keathley:
I'm wondering what a configuration for this might start to look like as we add more configurable values. There might be a good way to codify some of that in the future with but I'm not sure what it looks like yet. Just some stuff I started thinking about.

LGTM in any case.

PR #4, opened 2016-03-17
Wait for elements to be visible

alexwdaniel:
The max wait time should probably be configurable...

### DIFF
--- a/lib/wallaby/dsl/finders.ex
+++ b/lib/wallaby/dsl/finders.ex
@@ -27,7 +27,7 @@ defmodule Wallaby.DSL.Finders do
     do_find_elements(session, to_params(query))
   end
 
-  def any?(session, query) do
+  def any(session, query) do
     retry fn ->
       case do_find_elements(session, to_params(query)) do
         elements when length(elements) > 0 -> elements
--- a/lib/wallaby/dsl/finders.ex
+++ b/lib/wallaby/dsl/finders.ex
@@ -7,6 +7,7 @@ defmodule Wallaby.DSL.Finders do
   def find(session, query, [{:count, count} | _] = _opts) do
     retry fn ->
       case do_find_elements(session, to_params(query)) do
+        elements when length(elements) > 0 and count == :any -> elements
         elements when length(elements) == count -> elements
         [] -> raise Wallaby.ElementNotFound, message: "Could not find element"
         elements -> raise Wallaby.AmbiguousMatch, message: "Ambiguous match, found #{length(elements)}"
@@ -25,15 +26,6 @@ defmodule Wallaby.DSL.Finders do
 
   def all(session, query) do
     do_find_elements(session, to_params(query))
-  end
-
-  def any(session, query) do
-    retry fn ->
-      case do_find_elements(session, to_params(query)) do
-        elements when length(elements) > 0 -> elements
-        [] -> raise Wallaby.ElementNotFound, message: "Could not find any elements"
-      end
-    end
   end
 
   defp to_params({:xpath, xpath}) do
--- a/lib/wallaby/dsl/finders.ex
+++ b/lib/wallaby/dsl/finders.ex
@@ -1,6 +1,8 @@
 defmodule Wallaby.DSL.Finders do
   alias Wallaby.Session
 
+  @default_max_wait_time 5_000
+
   def find(session, query) do
     case do_find_elements(session, to_params(query)) do
       [element] -> element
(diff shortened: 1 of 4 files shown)


lib/wallaby/dsl/finders.ex
```diff
@@ -13,6 +15,17 @@ defmodule Wallaby.DSL.Finders do
     do_find_elements(session, to_params(query))
   end
 
+  def wait_until_visible(session, query, opts \\ [])
+  def wait_until_visible(session, query, [{:count, count} | _] = _opts) do
+    retry fn ->
+      case all(session, query) do
+        elements when length(elements) == count -> elements
```
alexwdaniel: It'd also be nice to wait for all without specifying a count. Maybe create a separate `wait_until_all_visible` or something? or have `wait_until_visible(session, ".main", count: :all)`. What do you think?

keathley: Wouldn't `count: :any` be equivalent to `count: 1` though?

keathley:
Do we want to use this implicitly inside of the finders as well? Or should we always expect people to do explicit waiting? I'm fine with just saying that for now it needs to be explicit and then maybe adopting the implicit waiting later if it makes sense.

lib/wallaby/dsl/finders.ex
alexwdaniel: I was thinking count: :any would wait until one or more elements were visible, or timeout. The use case I ran into was checking that an employee was on the bench. Depending on the size of the window, a different number of weeks are visible. So if I do `wait_until_visible(".bench__availability", count: :any)` I get some variable number of nodes. Alternatively, a more specific selector would solve my problem, but I'm not sure what that'd look like. I can't do `.bench__availability:first`. Maybe an xpath? Haven't used those before

alexwdaniel:
Chatted with Christian this morning... he made the case for waiting inside the normal finders as it reads more naturally and it separates knowledge of how the front-end works. I've been convinced but still wish waiting wasn't necessary

keathley:
It would be fine with me to expose the waiting helper and just use it internally inside the finder as well. That way you can be more explicit if you choose.

lib/wallaby/dsl/finders.ex
keathley: I'm surprised that you can't do `.bench__availability:first`. You could definitely use xpath for it but I'd really like to avoid having that be a use case for people. I could see value in a `count: :any` but `:any` to me means that it could also be 0. Is that just me?

alexwdaniel:
Cool. How do you imagine :count works? My interpretation is 'find this exact number or fail'. `all` needs to raise an error to trigger the retry...so with a specific count, find count or raise an error. Without a count, should it find at least one or raise? As it's implemented now, all can return an empty list

keathley:
I would imagine that in the case of `all` it just finds all the matches on the page currently. If you want to wait until you have a specific amount then you would use `find('#some-selector, count: 3)` or whatever. If it never finds that many then it throws the error. I don't think that `all` should find an error.  It definitely brings up the issue of consistency though. Is it semantically OK for all to just return you an empty list in that scenario?

alexwdaniel:
you're thinking `all` doesn't have the wait_until behavior? there could be a separate `any` function that keeps checking until at least one node is found

keathley:
Yeah I don't think we can wait for anything in `all`. Maybe `any` should be a matcher? So you would chain things like so:

```elixir
session =
  |> visit("/some_page.html")

assert any?(session, '.users')
```

What do you think about something like that?

lib/wallaby/dsl/finders.ex
```diff
@@ -1,11 +1,25 @@
...
   alias Wallaby.Session
 
-  def find(session, query) do
-    case do_find_elements(session, to_params(query)) do
-      [element] -> element
-      [] -> raise Wallaby.ElementNotFound, message: "Could not find element"
-      elements -> raise Wallaby.AmbiguousMatch, message: "Ambiguous match, found #{length(elements)}"
+  @default_max_wait_time 3_000
+
+  def find(session, query, _opts \\ [])
```
alexwdaniel: @keathley seemed to make more sense to just remove the wait_until

keathley: I'm fine with that

lib/wallaby/dsl/finders.ex
```diff
@@ -13,6 +27,15 @@ defmodule Wallaby.DSL.Finders do
     do_find_elements(session, to_params(query))
   end
 
+  def any?(session, query) do
+    retry fn ->
+      case do_find_elements(session, to_params(query)) do
+        elements when length(elements) > 0 -> elements
```
keathley: I think that this should return a boolean instead of the elements if its a `?` function.

alexwdaniel: maybe no ? then. Or have an `any?` and `any`

keathley: My initial thought was to just have `find/2` and `all/2` as finders for nodes. If you want to specify a count then you could add that as well.

`any?/` would be specifically for assertions. 

That seems like its not what you're going for though no?

keathley: @tokafish do you have thoughts on this?

PR #5, opened 2016-03-18
Debug helpers

alexwdaniel:


test/wallaby/dsl_test.exs
```diff
@@ -166,4 +166,24 @@ defmodule Wallaby.DSLTest do
...
   end
+
+  test "taking a screenshot", %{session: session, server: server} do
+    path =
+      session
+      |> visit(server.base_url)
+      |> take_screenshot
+
+    assert File.exists? path
+    File.rm_rf! path
```
alexwdaniel: It'd be nice to cleanup the /screenshots directory as well but I wasn't sure how to get the app root

alexwdaniel: Oh it should just be File.cwd

lib/wallaby/dsl/helpers.ex
```diff
@@ -0,0 +1,43 @@
+defmodule Wallaby.DSL.Helpers do
+  alias Wallaby.Session
+
+  def take_screenshot(session) do
+    image_data =
+      Session.request(:get, "#{session.base_url}session/#{session.id}/screenshot")
```
keathley: It doesn't need to happen in this PR but we should really abstract this away somehow.

keathley:
This looks good to me overall. Should the screenshots directory be configurable in some way? Or does that probably not matter?

alexwdaniel:
It's configurable in hound. It might be nice, but i don't really care where they go as long as they're not in the project root. I think it's ok for now.

keathley:
I'm fine with it as is.

PR #12, opened 2016-03-29
Add execute script and key press actions

alexwdaniel:


lib/wallaby/helpers/keys.ex
```diff
@@ -0,0 +1,12 @@
+defmodule Wallaby.Helpers.KeyCodes do
+
+  def json(keys) do
```
alexwdaniel: @keathley this is cribbed from hound.. but i wonder if there's a better way to encode unicode characters as json

lib/wallaby/driver.ex
```diff
@@ -223,10 +256,11 @@ defmodule Wallaby.Driver do
     %{using: "css selector", value: css_selector}
   end
 
-  defp request(method, url, params \\ %{}) do
+  defp request(method, url, params \\ %{}, opts \\ []) do
     headers = [{"Content-Type", "text/json"}]
     body = case params do
       params when map_size(params) == 0 -> ""
+      params when [{:encode_json, false}] == opts -> params
```
alexwdaniel: is there a way to match a keyword list that might also include other values inside the guard? that's not possible right? you'd need another case statement or to pattern match on the function signature?

keathley: Thats correct. You can't execute arbitrary functions inside a guard. This might be a good place to use the `with` syntax but I'm leaning towards just moving the whole case statement into its own function that we can then pattern match against.

lib/wallaby/helpers/keys.ex
keathley: I'm OK with this. Which part don't you like?

alexwdaniel: having to explicitly escape `{\"value\": ...}` is annoying. it's fine though

keathley: We could probably do something to make that more generic in the future. But given how simple it is I'm not super concerned about it.

alexwdaniel:
@keathley any idea why this is failing on travis?

keathley:
The server process is failing to respond in 5 seconds. The driver is timing out and failing because of that. I'm not sure why the server process would be blocking for that long. It might be an issue with communicating with phantom, or with trying to get a phantom process from the worker pool.

bomatson:
Would someone mind rebasing this branch with master? We're working on a spec for Allocations that's requiring the use of both `has_no_css?` (which is in master) and `execute_script` (which is in this branch)

keathley:
@bomatson, @alexwdaniel Branch has been rebased on latest master. Caution if you have the branch locally. If the tests are good I'll merge it into master and deploy a new version.

PR #21, opened 2016-03-30
New feature: Support for Selecting an Option from a Select Box

eiseneker:
It's all in the name! Ask @bomatson or I if any questions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,7 +136,7 @@ fill_in(session, "#last_name_field", with: "Keathley")
 choose(session, "Radio Button 1")
 check(session, "Checkbox")
 uncheck(session, "Checkbox")
-select(session, "Option 1", from: "My Awesome Select")
+select(session, "My Awesome Select", option: "Option 1")
 click(session, "Some Button")
 ```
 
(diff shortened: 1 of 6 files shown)


keathley:
Thanks for adding this!

I was wondering about the merits of the `from: ` option. Does it make sense to use that in this context? Or does it make more sense to just pass in the select box as the first option to the select? That would keep more in line with our existing semantics and would allow you to compose finders with selections for nesting purposes. I'm thinking of something like this:

```elixir
page
|> find(".signup-form")
|> select("Home State", option: "Florida")
```
or
``` elixir
page
|> find(".signup-form")
|> find(".home_state_field")
|> select(option: "Florida")
```

The `option` key might be unneeded here but it matches semantically with `fill_in(some_selector, with: foo)`.

What do you think about this?

bomatson:
For sure! We were trying to emulate the behavior of `fill_in` as much as we could. We'll refactor this to match more closely with:

```
page
|> find(".signup-form")
|> select("Home State", option: "Massachusetts")
```

Does that work?

keathley:
Sounds good to me.

keathley:
This looks great! :tada:

PR #22, opened 2016-03-31
Add note about version of Ecto required

sevenseacat:
These sandbox methods aren't in the latest published version, 1.1.5.

(Can be updated once Ecto 2.0.0 final is out)


keathley:
Thats a good point. I'm in the process of making the number of phantoms configurable. Once thats done then we can have specific instructions on Ecto 1 (meaning turn concurrency to 1) and Ecto 2.

For now this looks good to me.

PR #30, opened 2016-04-01
Features/has text retry

eiseneker:
Here's a shot at introducing a ``has_text?`` method, which replaces ``has_content?``

The intent of this method is to have it behave like ``find/2``, which retries. This would allow us to use text as keyframes for state of a page, e.g.

```elixir
assert node |> has_text?('Record updated!')
```

Some implications:

- ``has_text?/2`` allows for partial matching; this would allow us for to search for text on a page, rather than having to know the node ahead of time
- ``has_text?/2`` raises a Wallaby.ExpectationNotMet error instead of returning false
- ``text/2`` is not altered. It just returns the text of the node at that time

### DIFF
--- a/lib/wallaby/node.ex
+++ b/lib/wallaby/node.ex
@@ -242,11 +242,11 @@ defmodule Wallaby.Node do
 
   def has_text?(%Node{}=node, text) when is_binary(text) do
     retry fn ->
-      not_found = Regex.run(~r/#{text}/, text(node)) |> is_nil
-      if not_found do
+      regex_results = Regex.run(~r/#{text}/, text(node))
+      if regex_results |> is_nil do
         raise Wallaby.ExpectationNotMet, "Text '#{text}' not found"
       end
-      !not_found
+      true
     end
   end
 
--- a/lib/wallaby/node.ex
+++ b/lib/wallaby/node.ex
@@ -238,10 +238,13 @@ defmodule Wallaby.Node do
   @doc """
   Matches the Node's content with the provided text.
   """
-  @spec has_content?(t, String.t) :: boolean()
+  @spec has_text?(t, String.t) :: boolean()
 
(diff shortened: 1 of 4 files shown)


keathley:
Thanks! Would you mind rebasing this off of master? Otherwise things look good to me.

eiseneker:
@keathley All set!

lib/wallaby/node.ex
```diff
@@ -238,10 +238,16 @@ defmodule Wallaby.Node do
...
 
-  def has_content?(%Node{}=node, text) when is_binary(text) do
-    text(node) == text
+  def has_text?(%Node{}=node, text) when is_binary(text) do
+    retry fn ->
+      regex_results = Regex.run(~r/#{text}/, text(node))
+      if regex_results |> is_nil do
+        raise Wallaby.ExpectationNotMet, "Text '#{text}' not found"
+      end
+      true
```
keathley: I think this might have gotten changed during the rebase. Shouldn't this be `regex_results`?

eiseneker: Yeah, we snuck this in just now. Just trying to make it a bit more readable... the intent is to always pass it if something was found, and always fail if it can't.

Maybe there's a better way to do this? We were really hoping for an Elixir equivalent to regex_results.present? ...

keathley: The only thing that I'm worried about here is that its non-standard to throw errors from a `?` method. So it might break some people's expectations. I should have thought that through earlier.

Ideally what I'd like to see is that it would retry but only return boolean values. That might require some refactors to the `retry/1` function, but I'm OK with that (it needs to be refactored anyway).

eiseneker:
@keathley Bumped -- now with ``assert_text`` and ``has_text?`` functions

keathley:
Thanks!

PR #45, opened 2016-04-30
Fix setup documentation

chrismcg:
I found I needed the `:sql_sandbox` config in my `config/test.exs` rather than the test helper. This is because the endpoint is compiled by the time the test helper is run and so doesn't see the setting. Without this I got the `(RuntimeError) cannot find ownership process ` error.


keathley:
Sorry about the confusion. Thanks for updating this!

PR #46, opened 2016-05-03
Implemented functions to get current url and current path using session

praveenperera:


lib/wallaby/session.ex
```diff
@@ -124,6 +124,22 @@ defmodule Wallaby.Session do
...
+  """
+  @spec get_current_url(t) :: String.t
+  def get_current_url(session) do
+     Driver.current_url(session)
+  end
+
+  @doc """
+  Gets the current url of the session.
+  """
+  @spec get_current_url(t) :: String.t
```
keathley: Can you update this to the correct spect? I think it should be `get_current_path`. Looks like the `@doc` needs to be updated as well.

keathley:
Thanks for this! I left a few comments. Just some simple updates to the function spec and docs. Otherwise it looks good to me.

praveenperera:
Oops my mistake with the typos, fixed.

keathley:
Looks great!

PR #49, opened 2016-05-06
Add shared config to AcceptanceCase in README

chrismcg:
Manually checking out connections currently doesn't work for things like making a query from a `FooChannel.join` callback or a background job so you can't run your tests async in these cases. Having the `unless tags[:async]` bit in the AcceptanceCase matches what Phoenix does and allows you to quickly toggle between async and sync when you run into this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,8 +87,11 @@ defmodule YourApp.AcceptanceCase do
     end
   end
 
-  setup _tags do
+  setup tags do
     :ok = Ecto.Adapters.SQL.Sandbox.checkout(YourApp.Repo)
+    unless tags[:async] do
+      Ecto.Adapters.SQL.Sandbox.mode(YourApp.Repo, {:shared, self()})
+    end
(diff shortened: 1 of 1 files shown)


keathley:
Thanks! This looks good to me.

I'm wondering now if it would make sense to provide some documentation on what tests can be async or have to be sync. I've also been thinking about a way to provide a default AcceptanceCase or generator for an AcceptanceCase to help people get up and running quicker.

chrismcg:
Yeah, some docs on when to use sync would help I think. I ran into the problem when my page under test opened up a websocket connection. The `UserChannel` in my app talks to the database to check if the user is authorized for the channel and that blew up with the ownership error until I switched back to sync. I think the plan is to fix that particular case in Phoenix but it still applies if your request uses a Task or GenServer that talks to the database.

dustinfarris:
FYI, I am seeing this as well in Phoenix 1.3.0-rc.0.  Disabling `async` makes it go away.

PR #50, opened 2016-05-07
allow to reliably click buttons via click_button - fixes #47

### LINKED ISSUE #47: Button presses via click("<button-name>") don't register
At least it didn't work for me, although my buttons had similar markup to the ones in `test/support/pages/forms.html`, then again there seems to be no test coverage for it either. 

I wrote an xpath that is able to find the possible range of buttons with inspiration from Capybara. 

An implementation similar to `click_link` (and as I use it) would be:
```
def press(session, button)
  button_xpath = ".//*[self::input | self::button][(./@type = 'submit' or ./@type = 'reset' or ./@type = 'button' or ./@type = 'image')][(((./@id = '#{button}' or ./@name = '#{button}' or ./@value = '#{button}' or ./@alt = '#{button}' or ./@title = '#{button}' or contains(normalize-space(string(.)), '#{button}'))))]"

  session
  |> Node.find({:xpath, button_xpath}
  |> Node.click
end
```

If desired, I can create a PR.

follmann:


keathley:
Looks great!

follmann:
Thanks!

PR #54, opened 2016-06-20
Add title function for getting page titles

keathley:
This provides the driver and user api for finding the title for the current page.

### DIFF
--- a/lib/wallaby/driver.ex
+++ b/lib/wallaby/driver.ex
@@ -23,7 +23,7 @@ defmodule Wallaby.Driver do
 
     capabilities = Wallaby.Phantom.capabilities(user_agent: user_agent)
     params = %{desiredCapabilities: capabilities}
-    
+
     response = request(:post, "#{base_url}session", params)
     session = %Wallaby.Session{base_url: base_url, id: response["sessionId"], server: server}
     {:ok, session}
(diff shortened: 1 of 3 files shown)


lib/wallaby/session.ex
```diff
@@ -153,6 +153,15 @@ defmodule Wallaby.Session do
   end
 
   @doc """
+  Gets the title for the current page
+  """
+  @spec title(t) :: String.t
+
+  def title(session) do
```
rickclare: I would suggest making the function name more specific (e.g. `get_page_title()`) to follow the naming of similar functions `get_window_size()`, `get_current_path()` and to indicate this does not operate on DOM nodes.

lib/wallaby/driver.ex
```diff
@@ -83,6 +83,14 @@ defmodule Wallaby.Driver do
   end
 
   @doc """
+  Gets the title of the current page.
+  """
+  def title(session) do
+    resp = request(:get, "#{session.base_url}session/#{session.id}/title")
```
rickclare: Can this (is this) function specifically target `<head><title>Text</title></head>` ?  

It's possible that a `<title>` tag can be placed anywhere in a HTML document (obviously not a clever thing for a HTML developer to do, but it is possible), but we are only interested in the title that the browser uses.

keathley: Yeah it would be better to say something like `page_title`. I'll update that today

PR #60, opened 2016-07-02
Update version in README to 0.7.0

adamniedzielski:
BTW I don't know whether there is a community solution for that, but every library seems to specify the version number in the README and it's quite easy to forget to update it in the README while doing a release.


keathley:
I appreciate this :+1:. I really wish there was a better way to generate these things. I've thought about writing a custom mix task to do a lot of this legwork for me but the pain hasn't been great enough to mess with it yet.

PR #65, opened 2016-07-08
Fix so select can work when locator is a label with id

Thau:
Hello, I noticed that Wallaby can't find options in a select if the locator is a label's text and that label's "for" points to the select id.

This PR adds a test for that case and fixes it :).


keathley:
This is great thanks!

:tada:

Thau:
Of course. I just did it.

keathley:
Sorry I actually meant something  like `git rebase upstream/master`. That way it just moves the base of your branch to the latest commit. I went ahead and cherry picked your commit onto a new branch and merged that so we can keep the history nice and clean. This is in master now :tada:

PR #68, opened 2016-07-11
Document usage of send_keys and KeyCodes

sorentwo:
I ran into the need to send an `enter` key as part of an acceptance test. There wasn't any documentation within `Session.send_keys/2` about how it worked or how keys should be defined. This is a basic attempt at adding documentation to highlight that key code conversion if a feature, and hopefully lead future devs to the right part of the source code.


keathley:
This is great thanks! Would you mind running a `git rebase` against master before I merge it in?

sorentwo:
No problem, rebased and force pushed.

keathley:
Thanks!

PR #69, opened 2016-07-12
Workaround to be able to click Phoenix Method links

### LINKED ISSUE #67: `click_link` does not work with `method: :delete` links
Hi,

I am trying to use Wallaby to test the user session flow, it's a Phoenix app. So for the sign out, the EEx looks like this:

``` EEx
<%= link "Sign Out", to: session_path(@conn, :delete), method: :delete %>
```

Unfortunately this doesn't appear to work. Use `take_screenshot` I can see that the link is not getting clicked.

If I change the implementation to either:

``` EEx
<%= form_tag session_path(@conn, :delete), method: :delete %>
  <%= submit "Sign Out", class: "btn btn-primary" %>
</form>

# use `click_on` in the test
```

Or:

``` EEx
<%= link "Sign Out", to: session_path(@conn, :delete) %>

# change route from :delete to :get
```

Then the test works.

Also, the Phoenix JS is included correctly, and JS is being executed in the test environment.

Is this a known issue, or have I somehow done this incorrectly?

Thau:
This is a workaround to fix #67. It works in my use cases, but might be a little too rough, so I recommend a bit of testing and thinking about edge cases I didn't.

### DIFF
--- a/lib/wallaby/driver.ex
+++ b/lib/wallaby/driver.ex
@@ -75,6 +75,14 @@ defmodule Wallaby.Driver do
   end
 
   @doc """
+  Clicks a link
+  """
+  def click_link(%Node{session: session, id: id}) do
+    request(:post, "#{session.base_url}session/#{session.id}/element/#{id}/click")
+    request(:post, "#{session.base_url}session/#{session.id}/element/#{id}/submit")
+  end
+
+  @doc """
   Gets the text for an element
   """
   def text(node) do
--- a/lib/wallaby/driver.ex
+++ b/lib/wallaby/driver.ex
@@ -38,17 +38,17 @@ defmodule Wallaby.Driver do
   def find_elements(%Session{base_url: base_url, id: id}=session, query) do
     request(:post, "#{base_url}session/#{id}/elements", to_params(query))
     |> Map.get("value")
-    |> Enum.map(&cast_as_node({session, &1}))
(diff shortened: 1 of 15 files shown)


fredwu:
Thanks @Thau, I can confirm that this PR works for my use case (a phoenix :delete link) too. 👍

keathley:
Thanks for this @Thau! I'd still like to do a bit of research into the root cause before I merge this in. I'm also worried that there might be issues with firing both the click and the submit. My fear is that this could cause some unforeseen consequences or disrupt other workflows. I'll play around with it and see what I can come up with.

Thau:
Of course, @keathley. This is just to help people who found this problem like I did. My theory is that, as the link has a unique PhantomJS id, firing two events should not do any damage because they both are fired in the same element, but that's what I meant by edge cases.

I'll try to find some more complex tests of things that might fail.

Thank you :)

lukerollans:
Confirming this works for my use-case, which is super basic

keathley:
I've researched this a bit more. Looks like the root cause is that phantom doesn't implement `matches` which phantom uses to determine if it needs to submit the parent form. I'm going to submit a patch to phoenix since its just as easy to use `getAttribute` and avoids this issue for all browsers. If the PR doesn't get merged then we can provide a polyfill for phantom or implement @Thau's workaround for the phoenix driver.

PR #73, opened 2016-08-03
Adds a phantomjs_args config option to allow arguments to be passed to phantomjs.

seangeo:
Pretty simple change but allows customisation of the phantomjs process that Wallaby creates.

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,6 +71,12 @@ Wallaby requires PhantomJS to work. You can install PhantomJS through NPM or you
 $ npm install -g phantomjs
 ```
 
+You can pass arguments to PhantomJS through the `phantomjs_args` config setting, e.g.:
+
+```elixir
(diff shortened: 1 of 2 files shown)


keathley:
This looks great :tada:.

Out of curiosity was there a specific use case that you needed this for?

seangeo:
Hi. Thanks. Yes, I was getting JavaScript errors specific to phantomjs, this allowed me to turn on the phantomjs logs. I've also used it in the past for disabling image loading for faster tests. 

Cheers

keathley:
Cool. Disabling images was something that we thought about making explicit since it can save so much time when running tests. But this should provide a good workaround for people. I'm trying to figure out a good way to capture JS errors during test runs as well. Hopefully that'll be coming soon-ish. Thanks for the feedback :+1:

PR #78, opened 2016-08-15
Adds support for HTML file input form fields

### LINKED ISSUE #75: File upload?
Has file uploading been implemented in Wallaby? I looked but didn't see anything that would match a file input in xpath.ex nor did I see anything in the docs to indicate support.

stevegraham:
Closes #75.

### DIFF
--- a/test/wallaby/dsl/actions/file_test.exs
+++ b/test/wallaby/dsl/actions/file_test.exs
@@ -1,0 +1,48 @@
+defmodule Wallaby.Actions.FileTest do
+  use Wallaby.SessionCase, async: true
+
+  setup %{server: server, session: session} do
+    page =
+      session
+      |> visit(server.base_url <> "forms.html")
+
+    {:ok, %{page: page}}
+  end
+
+  describe "attaching a file to a form" do
+    test "by name", %{page: page} do
+      page
+      |> attach_file("file", path: "test/support/fixtures/file.txt")
+
+      assert find(page, "#file_field") |> has_value?("C:\\fakepath\\file.txt")
+    end
+
+    test "by DOM ID", %{page: page} do
+      page
+      |> attach_file("file_field", path: "test/support/fixtures/file.txt")
+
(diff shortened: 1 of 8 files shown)


test/wallaby/dsl/actions/file_test.exs
```diff
@@ -0,0 +1,41 @@
...
+
+    test "by label", %{page: page} do
+      page
+      |> attach_file("File", path: "test/support/fixtures/file.txt")
+
+      assert find(page, "#file_field") |> has_value?("C:\\fakepath\\file.txt")
+    end
+  end
+
+  test "attaching a non-extant file does nothing", %{page: page} do
```
keathley: I think this should throw an error if it can't find the correct selector.

keathley:
This is looking great @stevegraham!

The only issue that I can see is that it doesn't throw an error if it can't find the selector. The fix would be to use `find_field` instead of `find` in `Query.file_field`. That should take care of handling all of the proper assertions (visibility, name, count, etc.). If you also don't mind adding a clause in `Wallaby.QueryError.method` for file fields that would be great.

test/wallaby/dsl/actions/file_test.exs
stevegraham: This test defines what happens when a bad path is given. Not sure how much use that test is but I figured it would be useful to define the behaviour.

stevegraham:
Sure thing. I'll fix up this PR now.

stevegraham:
Added the things you asked for and also a bad label test as I noticed all other field tests included one.

test/wallaby/dsl/actions/file_test.exs
keathley: Oh of course. I misread this. In the future it might make sense to do a check to make sure that the file exists and then throw an exception or a warning if it doesn't. For now this is great.

keathley:
This looks great! :tada:

stevegraham:
❤️

Issue #79, opened 2016-08-16
Support for getting page session source


steve0hh:
Was wondering if you guys are ok with adding an addition API to get the page source.

This would aid developers that would like to stick with their preferred DSL like [floki](https://github.com/philss/floki).

I already have the function working in my fork of this project but would need some guidance in writing the tests. (where to place them etc..)

Example of the proposed API:

```
iex(4)> session |> Wallaby.Driver.visit("http://www.example.com") |> Wallaby.Driver.source()

"<!DOCTYPE html><html><head>\n    <title>Example Domain</title>\n\n 
the page source of the session currently
</html>"
```

keathley:
Thanks @steve0hh. A PR for this would be great. If you wanna submit what you have I can take a look and we can talk through any missing details.

I'm also curious to know a little more about how you're using floki. Is it specific to your use case? Are there specific pieces of functionality that are missing from Wallaby? Or is it just easier to manage then Wallaby's DSL? Looks like a cool project!

steve0hh:
No problem. I am pretty new to the elixir.

I need to build a automation tool that does some scraping.

Was thinking if this was a good use case. 😄  however, there's some issue with phantomjs not being able to render some of the sites I need to scrape. I might need to use selenium.

keathley:
Are they sites that use a lot of JS? They may be using new JS that phantom doesn't understand (and throwing errors) or they may just be taking a long time to load. I'm not sure if floki is doing any blocking but my suspicion is that it isn't. You can try the same query with the Wallaby DSL first which will block until the element appears on the page and see if that fixes your issue.

steve0hh:
Yes. It's basically an entire page driven by emberjs. (My FYP project)

> You can try the same query with the Wallaby DSL first which will block until the element appears on the page and see if that fixes your issue

I tried it too. It seems to be the problem with `phantomjs` not wallaby. 

Things I did:
- Suppressed the javascript errors (Complaining about a module not being able to load)
- Set timeout to infinite
- Used the `find()` (Couldn't find element)
- Took a screenshot and found out that the javascript wasn't running correctly

Another reason why I wanted to have the `source` function, is also because I wanted to scrape the HTML in the nodes too. 

Spent an afternoon looking through your code and editing it and thought that you guys might be ok with a PR hence this thread. 😄

keathley:
I'd double check the error in your page. It could be a red herring but you might need a polyfill or something like that for phantom. Thanks for the PRs

Issue #74, opened 2016-08-08
HTTPoison timeout


stevegraham:
I'm having some tests fail intermittently because of a HTTPPoison timeout error. Would you be open to setting the HTTPoison timeout to `:infinity` as ExUnit.Case already has a configurable timeout with a default?

EDIT: Adding some more information.

- This only happens when running my Wallaby tests on their own, e.g. `mix test test/features` 
- It's always the first or second test.

This makes me think it's actually the first test running before everything is fully initialised (phantom?), so setting HTTPoison mike not be the best way to fix in your opinion.

keathley:
There are a couple things that could be causing this. It *shouldn't* be possible to checkout a phantom before its started. A more likely solution is that its the phoenix server booting up / building assets. I'm not opposed to setting a longer timeout for httpoison though. Its probably better to return a more meaningful error from Wallaby anyway. Either way it would be worth digging into a bit to see exactly whats going on.

aaronjensen:
We hit this too due to webpack building assets.

keathley:
Would you mind submitting a PR for this? I won't be able to get to it for a few days.

steve0hh:
I've had this error and made a fork of this project.

Basically i did the following:

```
  defp make_request(method, url, body) do
    headers = [{"Content-Type", "text/json"}]
    case HTTPoison.request(method, url, body, headers, [timeout: :infinity, recv_timeout: :infinity]) do
      {:ok, response} ->
        Poison.decode!(response.body)
      {:error, e} ->
        raise "There was an error calling: #{url} -> #{e.reason}"
    end
  end

```

Essential just setting both poison options : `[timeout: :infinity, recv_timeout: :infinity] `.

It solves the timeout issue.

If you guys are ok, I could submit a PR.

keathley:
Please do.

PR #81, opened 2016-08-16
Add get source feature

steve0hh:
PR for issue #79, support for getting the current page source

and

PR for issue #74, setting the recv_timeout to infinity

### DIFF
--- a/lib/wallaby.ex
+++ b/lib/wallaby.ex
@@ -46,8 +46,12 @@ defmodule Wallaby do
   end
 
   def js_errors? do
-    Application.get_env(:wallaby, :js_errors) || true
+    Application.get_env(:wallaby, :js_errors)
+    |> explicitly_set()
   end
+
+  defp explicitly_set(:false), do: false
+  defp explicitly_set(_), do: true
 
   defp poolboy_config do
     [name: {:local, @pool_name},
--- a/lib/wallaby/driver.ex
+++ b/lib/wallaby/driver.ex
@@ -266,7 +266,7 @@ defmodule Wallaby.Driver do
   @doc """
   Retrives the current page source from session
   """
-  def source(session) do
+  def page_source(session) do
     check_logs! session, fn ->
       response = request(:get, "#{session.url}/source")
       response["value"]
--- a/lib/wallaby/driver.ex
+++ b/lib/wallaby/driver.ex
@@ -256,6 +256,16 @@ defmodule Wallaby.Driver do
     resp["value"]
   end
 
+  @doc """
+  Retrives the current page source from session
+  """
+  def source(session) do
+    check_logs! session, fn ->
+      response = request(:get, "#{session.url}/source")
+      response["value"]
+    end
+  end
+
   defp window_handle(session) do
     check_logs! session, fn ->
       request(:get, "#{session.url}/window_handle")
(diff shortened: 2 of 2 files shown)


steve0hh:
f209199 also fixes a big where the config variable js_errors is cannot be set to false.

keathley:
Thanks. I'd like to separate these issues into their own branches and merges. I'll just cherry pick your commits for the httpoison timeout and the js_errors issue. That way we can focus on just the logic for getting page source in this PR :+1:

lib/wallaby.ex
```diff
@@ -46,9 +46,13 @@ defmodule Wallaby do
   end
 
   def js_errors? do
-    Application.get_env(:wallaby, :js_errors) || true
+    Application.get_env(:wallaby, :js_errors)
```
keathley: `Application.get_env/3` actually accepts a default (which is what I should have used in the first place). Thanks for catching my silly mistake.

steve0hh: Oh cool!

Didn't know that. Another new thing learnt today. 😃

keathley:
The other fixes have been merged into master. If you could rebase your branch so that its up to date with master that would be rad.

Otherwise, I think the only other thing that this PR needs is a test. The way that I would do it is to create a new session test like: `test/wallaby/session/page_source_test.exs`. Then I would write a very basic test like:

```elixir
test "page_source/1 retrieves the source of the current page" do
  source = 
    session
    |> visit(server.base_url <> "/index.html")
    |> page_source
  
  html_path = Path.absname("test/support/pages/index.html", Application.app_dir(:wallaby))
  assert source == File.read!(html_path)
end
```

You'll probably have to adjust that test to work in the real world. But it should be close. You'll also need to make sure that the test case is `async: false` since its interacting with the outside world. Let me know if you have any issues :+1:

lib/wallaby/driver.ex
```diff
@@ -256,6 +256,16 @@ defmodule Wallaby.Driver do
     resp["value"]
   end
 
+  @doc """
+  Retrives the current page source from session
+  """
+  def source(session) do
```
keathley: Lets change this to `page_source` since it can only ever be used to get the source of the page.

casellepete:
Thank you for your efforts on this.  
Looking forward to it.

keathley:
@steve0hh are you still interested in finishing this up? If not thats cool. I can take it from here. But I thought I would check first.

steve0hh:
Hey @keathley , sorry about the delay. 

Was busy with school and needed to figure out how to merge upstream with my own fork so there's no conflicts. 😄

opsb:
Would be great to get `page_source` merged in, seems like a simple PR and would be happy to do it myself. @steve0hh perhaps you could put a PR together for just https://github.com/keathley/wallaby/pull/81/files#diff-58dd4483c60c6311fba480cefc54fbc6R266 ?

keathley:
No worries @steve0hh. Typically the best way to update your branch is to do a [rebase](https://git-scm.com/docs/git-rebase) instead of a merge. You can also use rebase to "squash" commits together or re-order commits. It helps keep the commit log clean and linear. Thanks for the work on this :+1:

@opsb I'll just cherry-pick the commit and do some cleanup.

keathley:
This has been merged here: 6f0e6babbd7862f6546b7601f74225ca3ccc7d93

keathley:
Thanks @steve0hh :tada:

PR #84, opened 2016-08-17
Increases gen_server:call timeout to :infinity

stevegraham:
Increases timeout from 5000ms to :infinity. In
environments with a low pool size, e.g. CI workers
and lesser powered laptops, test failures due to a
delay > 5000ms in checking out a worker from the
pool frequently occur. As ExUnit has a default max
time for a test case to complete (60_000ms), it
makes sense to avoid these failures by allowing
a test to block until a worker is checked back in
to the pool.

### DIFF
--- a/lib/wallaby.ex
+++ b/lib/wallaby.ex
@@ -31,7 +31,7 @@ defmodule Wallaby do
   end
 
   def start_session(opts \\ []) do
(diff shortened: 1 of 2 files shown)


keathley:
Looks good :tada:

PR #93, opened 2016-08-25
Separate local storage per phantom

whw:
To better isolate tests, I thought it would be a good idea for each PhantomJS process to have its own local storage directory.

### DIFF
--- a/lib/wallaby/server.ex
+++ b/lib/wallaby/server.ex
@@ -5,6 +5,10 @@ defmodule Wallaby.Server do
     GenServer.start_link(__MODULE__, [])
   end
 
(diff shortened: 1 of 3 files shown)


aaronjensen:
@keathley Is this good to go? We're using this and it's a necessary thing if you want to support local storage and multiple threads.

keathley:
Yup this looks good to me. Thanks @whw! :tada:

keathley:
@aaronjensen Just cut a new release with these fixes. Lemme know if you have any issues with it.

aaronjensen:
That's great, thank you @keathley

PR #105, opened 2016-10-10
Add wait_until

opsb:
@keathley so it wasn't until I got to the documentation that nagging doubts I had about this approach gave me pause for thought...

Seeing as Wallaby already has a bunch of finder functions the logical endpoint for this approach is another set of functions that all do the same thing, except they return the original session instead of the session with the narrowed scope. 

Ideally this would be handled with currying e.g.

```
      session
        |> visit(server.base_url <> "wait.html")
        |> wait_until(find(".main"))
```

or

```
      session
        |> visit(server.base_url <> "wait.html")
        |> wait_until(link("Home"))
```

I'll try and get some feedback regarding design and see if there's a better approach.

### DIFF
--- a/lib/wallaby/dsl/actions.ex
+++ b/lib/wallaby/dsl/actions.ex
@@ -183,4 +183,23 @@ defmodule Wallaby.DSL.Actions do
 
     parent
   end
+
+  # @doc """
+  # Waits until the provided query is satisfied
+  #
+  # ## Examples
+  # ```
+  # session
+  #   |> click_on("Some Button that fires async action")
+  #   |> wait_until(& find(&1, ".result-of-async-action"))
+  # ```
+  # ```
+  # session
+  #   |> click_on("Some Button that fires async action")
+  #   |> wait_until(& link(&1, "Added by async action"))
+  # ```
+  def wait_until(parent, fun) do
+    fun.(parent)
+    parent
+  end
 end
--- a/test/support/pages/wait_until.html
+++ b/test/support/pages/wait_until.html
@@ -1,0 +1,23 @@
+<!DOCTYPE HTML>
+<html>
+  <head>
+    <title>Wait until ...</title>
+
+    <script type="text/html" id="template">
+      <div class="main">
+        <div class="orange">orange</div>
+        <a id="link" href="#">a link</a>
+      </div>
+    </script>
+
+    <script type="text/javascript">
+      setTimeout(function() {
+        var template = document.getElementById("template").innerHTML;
+        document.body.innerHTML = template;
+      }, 700)
+    </script>
+  </head>
+
+  <body>
+  </body>
+</html>
--- a/test/wallaby/dsl/actions/wait_until_test.exs
+++ b/test/wallaby/dsl/actions/wait_until_test.exs
@@ -1,0 +1,27 @@
+defmodule Wallaby.Actions.WaitUntilTest do
+  use Wallaby.SessionCase, async: true
+  use Wallaby.DSL
+  alias Wallaby.Node.Query
+
+  test "wait until `find` is satisfied", %{server: server, session: session} do
+    updated_session =
+      session
+        |> visit(server.base_url <> "wait_until.html")
+        |> wait_until(& find(&1, ".main"))
+
+    assert has_css?(updated_session, ".main"), "waited until selector became present"
+    assert has_css?(updated_session, "html"), "has not narrowed scope"
+  end
+
+  test "wait until `Query.link` is satisfied", %{server: server, session: session} do
+    updated_session =
+      session
+        |> visit(server.base_url <> "wait_until.html")
+        |> wait_until(& Query.link(&1, "a", text: "a link"))
+
+    html_root = updated_session |> find("body")
+
+    assert has_text?(html_root, "a link"), "waited until text became present"
+    assert has_css?(updated_session, "html"), "has not narrowed scope"
+  end
+end


opsb:
Discussed with @martinsvalin [on slack some](https://elixir-lang.slack.com/archives/general/p1476094225019942), came to the conclusion that on balance the best API is:

```
session
        |> visit(server.base_url <> "wait.html")
        |> wait_until(& find(&1, ".main"))
```

let me know if you want to go in a different direction though and I'll amend accordingly.

opsb:
So I'm feeling my inexperience with Wallaby now... I've included an example that uses `Query.link` but I don't see it mentioned in the documentation. Is `Wallaby.Node.Query` intended for internal only use or general consumption? I'm starting to think having the more specific `wait_until(find, args)` is a better fit but I'll wait for your feedback.

keathley:
I think before we can solve this correctly some other refactors need to take place. I think its a mistake to continue creating custom queries for different types of selectors. In fact it was probably a mistake to implement it that way in the first place. Instead all of the queries functions should just return query data which we can then pass to a function like `find` or `wait_until`. That separation should enable us to do more with the query engine long term. It should also make the implementation for `wait_until` fairly trivial.

opsb:
@keathley yep, that sounds like a much better approach

aaronjensen:
For what it's worth, `wait_until` means something different in the capybara world, so it may create some cognitive dissonance. Granted, capybara removed `wait_until`, I believe, but what it used to do was poll until something was true/stopped throwing exceptions. In this way you could do something like wait until a url was something you'd expect it to be or wait until there were 4 buttons on a page--things that are hard to do w/ just selectors.

PragTob:
dito what @aaronjensen - it's a very common test helper we still use in application level test that catches RSpect expectation not met and other errors and waits for a time (for isntance checking that something has been created in the database). Another name might be better :)

keathley:
What would you suggest as an alternative name? Does this functionality still make sense semantically?

aaronjensen:
A name that kind of describes what it's doing is `fork`. More in this domain is maybe something like `check_for`?

That said, the pipe breaking operators are all pretty awkward to use in wallaby in general. I end up writing lots of functions that do something with the session then return the session. 

Maybe a bigger change could work:

```elixir
session
|> visit('/foo')
|> find('#bar', fn bar -> click(bar))
# and...
|> find('#baz', &click/1)
```

The theme is that all wallaby functions always return the session. If you want to do something with the thing you're finding, you provide a callback. Of course, this is a big, non-backwards-compatible change, but I think it'd increase the ergonomics quite a bit.

PragTob:
Reading this again I noticed that the helper we have is named differently (gladly my last weeks were spent in elixir or mostly unit tests) - what I as talking about is named `patiently` . 

In that light, `wait_until` seems good to me. As a personal not, I think `wait_for` could be slightly better. I.e. for me `until` implies a generic block/activity that I can wait for (from old-school generic until loops) while `wait_for` for me says wait for this thing, as in this query or this element.

Not sure, and the difference is very minuscule.

keathley:
I'm a fan of the name `wait_for`.

@aaronjensen I've thought about this a bit. I agree that returning the session is cumbersome some times. I'm not sure that using anonymous functions is the right approach though. Perhaps it is and I should give it some more thought. Can you post an example where having to pass sessions is awkward? It would be useful to talk about real code.

I tend to use `find` as a way to scope queries. At one point I wanted to add a `scope_to` helper that accepted a query. That way its obvious what you're trying to do. With the new query engine I've been working on I'm more inclined to work out a way to compose queries and eliminate the need for an explicit call to `find`.

For instance if you wanted to interact with a form:

```elixir
session
|> visit("/")
|> find(css(".some-form"))
|> fill_in(text_field("Name"), with: "Chris")
```

You could instead just do something like:

```elixir
form = css(".some-form")
name_field = magic_composition_api(form, text_field("Name"))

session
|> visit("/")
|> fill_in(name_field, with: "Chris")
```

I don't have that api worked out yet (obviously). But that might be a nice way to eliminate excess noise from tests.

aaronjensen:
Here are a couple examples:

```
  def check_checkbox(session, name) do
    find(session, "##{name}")
    |> check

    session
  end

  def range_select(session, id, options) do
    range_wrapper = find(session, "##{id}")

    click_button(range_wrapper, options[:first])

    if options[:last] do
      click_button(range_wrapper, options[:last])
    end

    session
  end

  def async_dropdown_select(session, name, options) do
    select_wrapper = find(session, "##{name}-wrapper")

    select_wrapper
    |> find(".Select-input input")
    |> fill_in(with: options[:query])


    select_wrapper
    |> find(".Select-option", text: options[:with])
    |> click

    session
  end

  def dropdown_select(session, name, options) do
    select_wrapper = find(session, "##{name}-wrapper")

    select_wrapper
    |> find(".Select-input")
    |> click

    select_wrapper
    |> find(".Select-option", text: options[:with])
    |> click

    session
  end

  def fill_in_dropdown_with_new_option(session, name, options) do
    select_wrapper = find(session, "##{name}-wrapper")

    # Add a newline to select the option we've just created
    new_option_text = "#{options[:with]}\n"

    select_wrapper
    |> find(".Select-input input")
    |> click

    select_wrapper
    |> find(".Select-input input")
    |> fill_in([{:with, new_option_text}])

    session
  end
```

and a few converted examples:

```elixir
  def check_checkbox(session, name) do
    find(session, "##{name}", &check/1)
  end

  def range_select(session, id, options) do
    find(session, "##{id}" fn range_wrapper ->
      click_button(range_wrapper, options[:first])

      if options[:last] do
        click_button(range_wrapper, options[:last])
      end
    end)
  end

  def async_dropdown_select(session, name, options) do
    find(session, "##{name}-wrapper", fn select_wrapper ->
      find(select_wrapper, ".Select-input input", &fill_in(&1, with: ))
      find(select_wrapper, ".Select-option", text: options[:with], &click/1)
    end)

    # or

    session
    |> find("##{name}-wrapper .Select-input input", &fill_in(&1, with: ))
    |> find("##{name}-wrapper .Select-option", text: options[:with], &click/1)
  end
```

keathley:
Just wanted to see what other people thought about this PR now that the new api is out. Is `wait_for` something that we still need? Or is that functionality covered with the new api? Does `assert_has` and `refute_has` solve this problem?

I think that this could be a useful piece of functionality but I also haven't found myself needing explicit blocking mechanisms since moving all of our tests to the new api. I'm also on the fence about including a function thats only goal is to try to control time. Curious to see what others think at this point.

PragTob:
For the basic case an `assert_has` does the same thing, so I don't see myself needing it :)

In our rails app, however, we sometimes use a more general `wait_until` function, that for instance waits until a record shows up in the database, like (from memory):

```
patiently do
  expect(Shipment.find_by(reference_code: "FOO-BAR")).not_to be_nil
end
```

It captures specific exceptions and then retries after n milliseconds for a maximum of m seconds.

But I guess, this is not something that should go into Wallaby, but is more of a general purpose thing.

keathley:
I'm going to close this for now since I think this specific function is being covered by `assert_has` and the new query api. Thanks for all the discussion everyone. It really helped to figure out the correct api direction 👍

viniciussbs:
Hi, folks.

I'm using `refute_has` in my tests and they were taking longer to finish because it was waiting 5 seconds to conclude that the element does not exist. This was solved reducing the `max_wait_time` to `500`. Until we start testing some pages with AJAX requests. The final solution was increase the `max_wait_time`.

If this `wait_until` accepts the waiting time as an argument, I could have a global max waiting time of 500ms again and use `wait_until` in specific tests. Another possible solution would be `refute_has` accepting `max_wait_time` as an argument.

Increasing the max waiting time globally is good when you are running your tests in a slow machine. In a project with a larger team I've configured it to accept an env var `WALLABY_MAX_WAIT_TIME` because we have devs with very powerfull machines and others with 7 years old MacBook Pro. But we don't have e solution to increase the waiting time for specific tests or specific checking.

I can open another issue if `wait_for` is not the perfect fit for my issue.

keathley:
I think it would be good to avoid `wait_until` if we can since it starts to break our rules about forcing more queries and actions to be declarative. I actually think that the issue is in `refute_has`. My experience has been that `refute_has` is almost always better written as a `assert_has` and flipping the query. So if you had something like `refute_has(css(".foo"))` it would be come `assert_has(css("foo", hidden: true))` or `assert_has(css("foo", count: 0))`. What do you think about that @viniciussbs?

viniciussbs:
Yep, the problem for me is in `refute_has`. But I can do something like `assert_has(css("foo", hidden: true))` because, in my case, the elements are removed from the DOM or never rendered.

PR #106, opened 2016-10-11
Update key_codes.ex

opsb:
Correct typo on :space keycode. I double checked the keycode against https://seleniumhq.github.io/selenium/docs/api/py/webdriver/selenium.webdriver.common.keys.html


keathley:
Thanks @opsb :tada:

PR #126, opened 2017-01-06
note about location/file type of AcceptanceCase

Ch4s3:


Ch4s3:
@keathley ping

PragTob:
Not sure, it needs the `note` as the sample path already says `.ex` and if you are in a Phoenix project already there's also model_case/conn_case and friends lying around there :)

keathley:
I think I agree with @PragTob. I'm not sure that we need to explicitly state this. Really, I would rather automate the generation of an `acceptance_case.ex` for people so we can simply avoid these types of issues. I'm going to close this for now. @Ch4s3 if you wanna help with automating this task via. a mix task let me know and we can discuss how that might be implemented.

Ch4s3:
I'm a little fuzzy about why I wanted this in the readme. I'll take a look at automating it.

PR #127, opened 2017-01-06
Fix parentheses warnings for Elixir 1.4

xtian:
Elixir 1.4 gives warnings like the following:

```
warning: variable "driver" does not exist and is being expanded to "driver()",
please use parentheses to remove the ambiguity or change the variable name
  lib/wallaby.ex:25
```


keathley:
Looks great :tada:

PR #130, opened 2017-01-17
Refactor Queries and Actions.

keathley:
## Where we are

The original intention of the Wallaby api was to give ruby developers a familiar api to Capybara and to create human readable tests. This is an example of such a test:

```elixir
test "some web page", %{session: session} do
  session
  |> visit("some_page.html")
  |> find(".a-form")
  |> fill_in("Name", with: "Chris")
  |> click_button("Submit")

  assert session
  |> find(".header")
  |> text == "Welcome Chris!"
end
```

This api is certainly friendly but I believe it has a lot of drawbacks. The first issue is that its not consistent. `find` accepts css while other actions such as `fill_in` accept either label text, the name of the field, or the id of the element. Using the id tends to be confusing because you don't use css selector like `#some-input` you use the id value like `some-input`. Different actions have different query semantics. These issues require Developers to learn each actions semantics and constricts their ability to solve problems when a use case deviates from convention.

A deeper problem, and one that affects me as a maintainer, is that actions are coupled to queries. This means that if we want to update or add functionality to the query engine, we have to make sure that it works with *all* of the actions. Updating the query engine to find elements by text or to find a certain number of elements becomes difficult to test in each action. Its also why we have multiple ways of doing the same thing. For instance `click_button` and `click_link` do effectively the same thing. They click elements. But because both of those actions are coupled to querying we need to support 2 different functions.

A byproduct of this tight coupling is that test code is much harder to re-use. In practice I've found my tests to be brittle and I've struggled to combat that with the current api.

## Where I want us to go

I believe that a necessary step to grow and maintain Wallaby is to separate queries from actions. This PR is a step towards that. For instance the test above would now look like this:

```elixir
test "some web page", %{session: session} do
  session
  |> visit("some_page.html")
  |> find(css(".a-form"))
  |> fill_in(text_field("Name"), with: "Chris")
  |> click(button("Submit"))

  assert session
  |> find(css(".header"))
  |> text == "Welcome Chris!"
end
```

This change may seem verbose but it actually opens us up to writing much more maintainable tests. for instance we could extract each of those queries to an attribute or function call. That way when our presentation layer changes we only need to update our test code in a single location.

Any other query parameters will now be part of the query itself.

```elixir
css(".some-css", text: "With this text", count: 3)
css(".some-hidden-css", visible: false)

radio_button("My fancy radio button")
```

By extracting the query logic from our actions we can drastically reduce the footprint of Action api. Actions will accept queries. That means you can use css queries with `fill_in` or textfield queries with `find`, etc.

All of the action logic has been moved to a new module called `Browser`. The `Browser` is the entry point for interacting with, you guessed it, the browser. If you've used Ecto this pattern will probably be familiar to you.

## How are we gunna get there?

I'm keeping everything backwards compatible for now. The existing query actions (which I refer to as "extended actions") will be moved to their own module. The release after this release (which will hopefully be the 1.0 release) will make the extended actions opt-in.

There are still a few things to finish up but you should be able to start trying out the new functionality in your existing applications. Please try this out and let me know what you think. I'd really appreciate the feedback.

## Todos

- [x] Finish adding tests and documentation
- [ ] Add deprecation notices to extended actions.
- [ ] Move extended actions to their own module
- [x] Update Wallaby::DSL
- [x] Rename `StatelessQuery` to `Query`
- [ ] Update README
- [x] Rebase and squash commits


PragTob:
I like this API change/new API. As you note, it'll hopefully get people to extract methods for finding specific things which will lead to DRYer tests and it's what I do when writing rspec feature specs anyhow (mostly encapsulated in something like page objects there), so I think it's a good change.

I also think it's a good change as query and action should not be mixed as you mentioned as well, they are 2 different concerns. While capybara's API is very convenient that also leads to some shortcomings and separating these 2 different things is more in the spirit of elixir, or so I believe.

So, big :+1: from me!

lib/wallaby/query/error_message.ex
```diff
@@ -0,0 +1,173 @@
...
+    Text is based on visible text on the page. This is a limitation of webdriver.
+    Since the element isn't visible the text isn't visible. Because of that I
+    can't apply both filters correctly.
+    """
+  end
+  def message(_, :min_max) do
+    """
+    The query is invalid. Cannot set the minimum greater than the maximum.
+    """
+  end
```
PragTob: I like this implementation!

lib/wallaby/stateless_query.ex
```diff
@@ -0,0 +1,278 @@
...
+
+  ```
+  visit("/page.html")
+  |> find("#main-page .dashboard")
+  ```
+
+  If more complex querying is needed then its possible to use XPath:
+
+  ```
+  find(page, {:xpath, "//input"})
```
PragTob: module doc mentions `find` all the time alhough it doesn't seem to be defines here anymore, or well more narrow scoped methods like `css`/`xpath` which I like better than passing a tuple to find either way :)

mix.exs
```diff
@@ -33,6 +33,7 @@ defmodule Wallaby.Mixfile do
       {:earmark, "~> 0.2", only: :dev},
       {:ex_doc, "~> 0.11", only: :dev},
       {:issuer, "~> 0.1.2", only: :dev},
+      {:quixir, "~> 0.9.0", only: [:dev, :test]},
```
PragTob: :+1:

lib/wallaby/browser.ex
```diff
@@ -0,0 +1,831 @@
...
+  def page_source(session) do
+    {:ok, source} = Driver.page_source(session)
+    source
+  end
+
+  @doc """
+  Sets the value of an element.
+  """
+  def set_value(element, value) do
+    {:ok, _} = Driver.set_value(element, value)
```
keathley: I think it makes sense to return the element here so actions can be chained. All of the explicit pattern matches to `{:ok, _}` need to be turned into case statements so that the browser handles errors correctly. Otherwise people could get internal wallaby errors thrown in their tests.

lib/wallaby/browser.ex
```diff
@@ -0,0 +1,831 @@
...
+  """
+  @spec check(Element.t) :: Element.t
+  @spec check(parent, StatelessQuery.t) :: parent
+  @spec check(parent, locator, opts) :: parent
+
+  def check(%Element{}=element) do
+    cond do
+      checked?(element) -> element
+      true -> click(element)
+    end
```
keathley: I tend to use `cond` over case if the function is going to return a boolean. I don't use `if/else` just in general as `cond` achieves the same thing and is easier to extend down the road.

keathley:
Thanks @PragTob. I would definitely love some help writing tests for the existing actions, updating documentation, making sure that the browser operations are chainable, etc. I think the last "big" thing is to add some deprecation warnings to the existing actions and to make sure that the actions handle errors from the driver correctly. Specifically we need to be matching on more then just `{:ok, result}` in the actions. Going forward the driver shouldn't be responsible for throwing errors. The `Browser` should be doing that.

keathley:
I'm going to go ahead and merge in what I have now. I think we're at a point where we can work on the remaining issues in parallel. I'll create issues in case others want to contribute.

PragTob:
:tada: 

I'll see if I can really dig up the time, as my the first part of my simplecov work is done I might even.

Thanks a lot! :)

shannonwells:
One of the things that has always bugged me about Capybara is exactly that inconsistency you mention. Because of it, even after using Capybara for a few years, I still forget sometimes which type of specifier a method requires and I have to look it up.  I don't mind at all that it is mildly more verbose; by being more clear and consistent, it also helps devs who are new to the test framework.  So 👍 👍  for the API choice.

PR #132, opened 2017-02-06
defdelegate instead of ugly function

asiniy:


keathley:
This has been fixed upstream

PR #135, opened 2017-02-14
Introduce find with callback function.

keathley:
Currently chaining is one of the more tedious pieces of the api. It effects the maintenance of the external api and is one of the major sources of bugs. I break this all the time, and when it happens the errors tend to manifest themselves *very* far away from the actual issue. It also presents a challenge for end users. One of my goals is for Wallaby's api to encourage users to write functions, page modules, and other small helpers for creating re-usable tests. Currently chaining isn't _convenient_ and users have to go out of their way to remember to make each of their helpers chainable.

This change introduces a new `find` function: `find/3`. This new find function accepts an anonymous function as its last argument. The type signature looks like this:
`find(parent, Query.t, ((Element.t) -> any())) :: parent`.

Because the parent is returned from find rather then returning an element, finds can be chained together without confusion about what is being returned. It also allows us to create a better divide between the Element api and the Browser api. Now we can create a very limited set of functions that are specific to elements and build more complex actions using `find`. If users want to build custom actions then they can use these same low level actions and they won't have to worry about making their custom actions chainable. I believe that this change will also encourage users to write custom functions and pass those to `find` the same way that we do with `each` or `map`. For instance if you wanted to fill in a form it might look something like: `find(Query.css(".my-form"), &sign_up/1)`

This PR isn't complete yet. I wanted to get some feedback before I went much farther. I also wanted to see what others thought about creating a smaller api for elements and what actions we could move. I'm also curious to see what others think about the older api. For now I've implemented actions such as `fill_in` and `click_button` using the new `find` and basic operations on elements. Should we retain those functions? Or does it make sense to always `find` and then `operate` on an element? Personally I still prefer to read tests such as:

```
session
|> fill_in(Query.text_field("Name"), with: "Chris")
|> click(Query.button("Save"))
```
as opposed to:

```
session
|> find(Query.text_field("Name"), &(fill_in(&1, with: "Chris")))
|> find(Query.button("Save"), &click/1)
```

But I would like to see what others think. Personally I think any confusion is easily solved by creating both a rich api in the `Browser` module and a small more focused api in the `Element` module. I'm comfortable moving some of that functionality out of `Browser` again now that there is a better path forward and less separation between `Element` (formerly `Node`), `Session`, and `DSL`. One of my plans is to remove the `import Wallaby.DSL` from the `feature_case` and instead replace it with an `alias`. That should also help users understand the different types that they are dealing with.


bunnymatic:
i definitely like where you want to go (as written up in here https://github.com/keathley/wallaby/pull/130) and I think this idea with the query is not bad.

I also dislike that capybara methods that for the different methods that take a `locator`, not all of them do the same thing (find by string, or by id or ...).  So being clear about the element with a Query thing seems good.

And in the thing you have written up in the other PR, it seems like this `Query.text_field` might later become `text_field` which would make this even cleaner.  

I say press on.

In terms of backwards compatibility, is there an easy way to report deprecations for those folks using wallaby?  so they have a chance to update their code, and you have a chance to really clean out old stuff?

keathley:
@bunnymatic I'll be adding deprecation warnings with code snippets so people can start updating in the next version.

opsb:
@keathley this looks like a huge improvement 💯

PR #147, opened 2017-02-27
Inspection in deprecation warnings

keathley:
Opts could throw an error when used in the deprecation warnings. Added an `inspect` to make sure that everything works correctly.


PragTob:
:+1:

PR #154, opened 2017-03-10
quote arguments to script, use spawn_executable

schnittchen:
Wallaby was unable so spawn the subprocess when the project path contains a space.
Additional quoting added by using the args option.

### DIFF
--- a/lib/wallaby/phantom/server.ex
+++ b/lib/wallaby/phantom/server.ex
@@ -68,7 +68,6 @@ defmodule Wallaby.Phantom.Server do
     Application.get_env(:wallaby, :phantomjs_args, "")
     |> case do
       string when is_binary(string) ->
-        # Backwards compatible. Should be deprecated?
         String.split(string)
       list when is_list(list) ->
         list
--- a/lib/wallaby/phantom/server.ex
+++ b/lib/wallaby/phantom/server.ex
@@ -25,13 +25,18 @@ defmodule Wallaby.Phantom.Server do
     port = find_available_port()
     local_storage = tmp_local_storage()
 
-    Port.open({:spawn, phantomjs_command(port, local_storage)}, [:binary, :stream, :use_stdio, :exit_status])
+    Port.open({:spawn_executable, script_path()},
+      [:binary, :stream, :use_stdio, :exit_status, args: script_args(port, local_storage)])
 
     {:ok, %{running: false, awaiting_url: [], base_url: "http://localhost:#{port}/", local_storage: local_storage}}
   end
 
-  def phantomjs_command(port, local_storage) do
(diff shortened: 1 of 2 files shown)


lib/wallaby/phantom/server.ex
```diff
@@ -61,6 +66,13 @@ defmodule Wallaby.Phantom.Server do
 
   defp args do
     Application.get_env(:wallaby, :phantomjs_args, "")
+    |> case do
+      string when is_binary(string) ->
+        # Backwards compatible. Should be deprecated?
```
keathley: We can leave the string handling for now so its fine to remove this comment.

lib/wallaby/phantom/server.ex
```diff
@@ -25,13 +25,18 @@ defmodule Wallaby.Phantom.Server do
...
+      [:binary, :stream, :use_stdio, :exit_status, args: script_args(port, local_storage)])
 
     {:ok, %{running: false, awaiting_url: [], base_url: "http://localhost:#{port}/", local_storage: local_storage}}
   end
 
-  def phantomjs_command(port, local_storage) do
-    "#{script_path()} #{phantomjs_path()} --webdriver=#{port} --local-storage-path=#{local_storage} #{args()}"
+  def script_args(port, local_storage) do
+    [
+      phantomjs_path(),
```
keathley: Now that we're using `:spawn_executable` we should think about updating `phantomjs_path()` to use the `System.find_executable/1`.

schnittchen: What would be the benefits of this?
(also, the conversion could happen inside the script)

schnittchen:
I had to touch a test checking that arguments are passed already: https://github.com/schnittchen/wallaby/blob/4acfde96a08873993648f579042ec747938066ac/test/wallaby/phantom/configure_test.exs#L18 What's missing here?

lib/wallaby/phantom/server.ex
keathley: ah, you're right. We pass in the full path to script_path so it shouldn't be an issue.

keathley:
Since we're changing the way that phantom args are being added to the list of script args it would be good to have a test for it. Just something that shows if the configuration includes phantom args they end up in `script_args()` correctly.

schnittchen:
Hold on. It's failing me locally...

schnittchen:
This should work reproducably, but it's a bit verbose, and does not deal with the case the env value has been explicitly set to nil. Maybe not much of an issue, but should a similar thing be needed in another test I'd extract the env fiddling and use fetch_env instead.

WDYT?

keathley:
LGTM. Thanks! :tada:

PR #156, opened 2017-03-12
Try naming screenshot with module/test name

alexwdaniel:
@keathley Here's a pretty naive approach to capturing the test name in screenshot file names. The filename will end up being pretty long if we include the full test module name, description, and test blocks.

Questions:
- should we truncate the file name some? maybe only include the describe/test blocks? Or even separate things into directories based on module/test?
- how to you feel about extending the sessions struct to include info on the text context?
- I was going to update the readme to document how to modify the acceptance case included in the user's application to capture the test context. Does it make sense to create a separate screenshot case for the wallaby tests so the default `test/support/session_case.ex` stays the same?


keathley:
I think separating into directories based on describe, test, etc. would be rad. We'll just have to make sure that you can still retrieve artifacts on circle, travis, etc.

I'm ok with extending the struct to include this info. This won't work if someone takes a screenshot of a specific element though correct? To do that I think we'd have to traverse back up the parent list until we got back to the session.

alexwdaniel:
Ah right.... I hadn't thought about the element case. I can take a look at that this weekend. 

How would a nested directory structure for screenshots complicate things on ci? Are there specific problems that you can think of?

keathley:
I just wanna make sure that CI environments will still capture the artifacts correctly. I don't think it'll be an issue. Just something that we should test.

ellery44:
Hi @alexwdaniel @keathley could you do the following to handle Element screenshot case?:

```
  defp path_for_screenshot(%Wallaby.Element{parent: session}) do
    File.mkdir_p!(screenshot_dir())
    "#{screenshot_dir()}/#{basename_for_screenshot(session)}#{:erlang.system_time}.png"
  end
  def path_for_screenshot(session) do
    File.mkdir_p!(screenshot_dir())
    "#{screenshot_dir()}/#{basename_for_screenshot(session)}#{:erlang.system_time}.png"
  end
```

Add right here:
https://github.com/keathley/wallaby/pull/156/files#diff-1ffba8548aa79aeac761e129a8f40654R1060

PragTob:
Heyho @alexwdaniel !

Thanks for all the work. A lot has changed on master so this is ridden with merge conflicts now. Any chance we can get this cleaned up and get it in for the looming 0.17 release? I knew I'd want to have this feature and Al and Nala agree.

PR #158, opened 2017-03-16
Take screenshots if `has?` fails.

keathley:
This allows us to take screenshots if a `has?` returns false. I'm wondering if this is a good idea since a user might be expecting a false value. For instance if they did a `refute has?(query)`. However, getting automatic screenshots when we fail assertions makes debugging, especially on CI servers, a lot easier.

### DIFF
--- a/lib/wallaby/browser.ex
+++ b/lib/wallaby/browser.ex
@@ -830,8 +830,11 @@ defmodule Wallaby.Browser do
 
   def has?(parent, query) do
     case execute_query(parent, query) do
-      {:ok, _} -> true
-      {:error, _} -> false
+      {:ok, _} ->
+	true
+      {:error, _} ->
+	take_screenshot(parent)
+	false
     end
   end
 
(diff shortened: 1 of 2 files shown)


knewter:
what about adding a `refute_has?` and `assert_has?` of your own?

keathley:
I was hoping to avoid adding specific assertions and instead rely on exunits default `assert`. Having our own would simplify a lot of other issues though. Phoenix provides specific assertions so maybe thats not a terrible pattern.

PragTob:
@keathley I'm currently writing version one of our wallaby powered acceptance tests, I'm planning on PRing general helpers I write but maybe we should start with that in an issue. I think they definitely have value as specific helpers allow to write them as part of the pipe (as I can just return the parent/session again)

keathley:
I'm going to close this for now in favor of the `assert_has?` and `refute_has?` helpers discussed in #159

PR #160, opened 2017-03-19
Implement assert_has and refute_has

### LINKED ISSUE #159: assert_has/refute_has helpers
This was briefly touched upon in #158 (brought up by @knewter ) - if we want to have some custom assert_* helpers in wallaby specifically `assert_has`/`refute_has`

I think it's a good idea and an easy implementation, I currently have my custom helpers like this:

```elixir
    defp assert_has(parent, query) do
      assert has?(parent, query)
      parent
    end
```

Among having an assert ready, it allows to write assertions nicely as part of a pipe by returning the parent element:

```elixir
      session
      |> visit("/")
      |> take_screenshot
      |> assert_has(@login_button)
      |> login_as(me)
      |> take_screenshot
      |> assert_has_no(@login_button)
      |> assert_has(@logout_button)
      |> assert_text(@page_content, "Welcome")

```

(note right now I still have one valled `assert_has_no` but realized it's a bad idea and will change to `refute_has`)

I wanna finish up my work tomorrow and then PR these helpers and maybe others I have over the weekend/discuss them here as I feel they might be beneficial for a lot of people :)

PragTob:
* fixes #159
* implemented as macros as it results in a cleaner stack trace

This is also the first time I write a macro for production code, so feel free to review away :)

----

As for macro vs. function I tried out a quick version in our project (different implementation than here, hence less cool error message) and the stack trace is just cleaner:

Function:

```
  1) test logging in logging successfully (Fulfillment.AuthAcceptanceTest)
     test/acceptance/auth_acceptance_test.exs:5
     Expected truthy, got false
     code: has?(parent, query)
     stacktrace:
       (fulfillment) lib/wallaby_helpers.ex:20: WallabyHelpers.assert_has/2
       test/acceptance/auth_acceptance_test.exs:12: (test)
```

Macro:

```
  1) test logging in logging successfully (Fulfillment.AuthAcceptanceTest)
     test/acceptance/auth_acceptance_test.exs:5
     Expected truthy, got false
     code: has?(parent, login_button())
     stacktrace:
       test/acceptance/auth_acceptance_test.exs:12: (test)
```

(it goes directly to the test without the intermediate step where the function is defined).

### DIFF
--- a/lib/wallaby/browser.ex
+++ b/lib/wallaby/browser.ex
@@ -921,10 +921,10 @@ defmodule Wallaby.Browser do
       parent = unquote(parent)
       query  = unquote(query)
 
-      case has?(parent, query) do
-        true ->
+      case execute_query(parent, query) do
+        {:ok, _query_result} ->
           parent
-        false ->
+        {:error, _not_found} ->
           raise Wallaby.ExpectationNotMet,
                 Query.ErrorMessage.message(query, :not_found)
 
--- a/lib/wallaby/browser.ex
+++ b/lib/wallaby/browser.ex
@@ -903,6 +903,66 @@ defmodule Wallaby.Browser do
   end
 
   @doc """
+  Checks if `query` is present within `parent` and raises if not found.
+
+  Returns the given `parent` if the assertion is correct so that it is easily
+  pipeable.
+
+  ## Examples
+
+      session
+      |> visit("/")
+      |> assert_has(Query.css(".login-button"))
+  """
+  @spec assert_has(parent, Query.t) :: parent
+
+  defmacro assert_has(parent, query) do
+    quote do
+      parent = unquote(parent)
+      query  = unquote(query)
+
+      case has?(parent, query) do
+        true ->
+          parent
+        false ->
+          raise Wallaby.ExpectationNotMet,
+                "Query type '#{query.method}' with selector " <>
+                  "'#{query.selector}' not found"
+      end
+    end
+  end
+
+  @doc """
+  Checks if `query` is not present within `parent` and raises if it is found.
+
+  Returns the given `parent` if the query is not found so that it is easily
+  pipeable.
+
+  ## Examples
+
+      session
+      |> visit("/")
+      |> refute_has(Query.css(".secret-admin-content"))
+  """
+  @spec refute_has(parent, Query.t) :: parent
+
+  defmacro refute_has(parent, query) do
+    quote do
+      parent = unquote(parent)
+      query  = unquote(query)
+
+      case has?(parent, query) do
+        false ->
+          parent
+        true ->
+          raise Wallaby.ExpectationNotMet,
+                "Query type '#{query.method}' with selector " <>
+                  "'#{query.selector}' found but expected not to be there"
+      end
+    end
+  end
+
+  @doc """
   Searches for CSS on the page.
   """
   @spec has_css?(parent, Query.t, String.t) :: boolean()
--- a/lib/wallaby/browser.ex
+++ b/lib/wallaby/browser.ex
@@ -924,7 +924,8 @@ defmodule Wallaby.Browser do
       case execute_query(parent, query) do
         {:ok, _query_result} ->
           parent
-        {:error, _not_found} ->
+        {:error, {:not_found, results}} ->
+          query = %Query{query | result: results}
           raise Wallaby.ExpectationNotMet,
                 Query.ErrorMessage.message(query, :not_found)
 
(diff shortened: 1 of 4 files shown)


keathley:
Thanks! Overall this looks good to me. I'll pull this branch down a little later so I can play around with it. I wonder what those stack traces are going to look like in long pipelines.

PragTob:
:+1:

For me, the stacktraces point to exactly the line where I used the macro which is great! =D

If we get this to a state that you like I'd be happy/cool with to adjust assert_text to have more of the same API (right now it returns true instead of parent and is not a macro iirc).

And maybe add an assert for a current_url, although we can discuss that in a separate issue :)

keathley:
a `assert_current_url` macro seems useful. But, I think that we should deprecate `assert_text` and `assert_css` in favor of using `assert_has` and queries.

PragTob:
Seems fair! Will gladly do that as well :)

keathley:
Lets do the deprecation and the other helper as a separate PR. I tried this locally and it looks really good! My only thought was that we could use the `Query.ErrorMessage` as the message returned to the user. We already have that logic there so its something we can take advantage of.

PragTob:
:+1: definitely, noticed these nice messages while I got some normal `has?`'s failing, will see how I can incorporate them.

Definitely, other stuff next PR!

PragTob:
I played around with it a bit, to use `ErrorMessage` I need to change it to use more low level methods than `has?` as I need to have the query returned with its result to generate proper error messages. Will get to it a bit later, latest this evening hopefully :)

keathley:
You shouldn't need to drop below `has?`. You can just raise the exception directly like: `raise Wallaby.QueryError, ErrorMessage.message(query, :not_found)`

PragTob:
@keathley for the error message to be correct, it needs the `result` key (especially for the `refute` case). E.g. (unadjusted but there should be a not in the beginning) it comes out as: `"Expected to find any, visible elements that matched the css '.user' but 0, visible elements were found.\n"` e.g. it would be expected not to find any visible elements and all of that is done on `query.result` so I'm fairly certain I need it there :)

PragTob:
So, the error messages now look like this:

```
"Expected to find 1, visible element that matched the css '.something-else' but 0, visible elements were found.\n"
# and
"Expected not to find any, visible elements that matched the css '.user' but 6, visible elements were found.\n"
```

There's a couple of things that I'd like to discuss/are not optimal choices right now probably.

lib/wallaby/browser.ex
```diff
@@ -903,6 +903,65 @@ defmodule Wallaby.Browser do
...
+
+  defmacro refute_has(parent, query) do
+    quote do
+      parent = unquote(parent)
+      query  = unquote(query)
+
+      case execute_query(parent, query) do
+        {:error, _not_found} ->
+          parent
+        {:ok, query} ->
```
PragTob: `execute_query` is used here to get the query with the result for a nice error message, above at `assert_has` I still use `has?` as I noticed that in the `error` clause of `execute_query` the second argument isn't the `query` so it doesn't exactly help there. At the same time, it breaks the symmetry between the two - not sure what to best do here. Input welcome :)

lib/wallaby/browser.ex
```diff
@@ -1043,7 +1102,7 @@ defmodule Wallaby.Browser do
     end
   end
 
-  defp execute_query(parent, query) do
+  def execute_query(parent, query) do
```
PragTob: as they are implemented as macros we'd need this to be public, which begs the question if it's really the right interface to use here (I just used what `has?` uses), if this should be public facing, or if we should make another version/wrapper that is public facing.

lib/wallaby/query/error_message.ex
```diff
@@ -57,6 +58,12 @@ defmodule Wallaby.Query.ErrorMessage do
     """
   end
 
+  defp found_error_message(query) do
```
PragTob: not sure that this is the best method name :)

lib/wallaby/browser.ex
keathley: Ah ok, I think I see the issue. Looks like we will need to use `execute_query` for both `assert_has` and `refute_has`. The error for `execute_query` should be something like `{:not_found, results}`. So you should be able to get the result value out of the error in `assert_has`. If not then there is probably a different bug somewhere.

lib/wallaby/browser.ex
keathley: I'm ok with making this public.

PragTob:
@keathley think/hope this is good to go now :)

PR #171, opened 2017-03-25
Introduce Basic Benchmarking with benchee

PragTob:
Gives us a base line for important actions to benchmark against,
so we can see how improvements pan out :)

Added in the HTML formatter as well as it helps with certain things, aka shows you the raw run times so you see that it doesn't just get slower or faster over time which is nice and generally I think it's nicer. Plus maybe I'll use wallaby to test the HTML formatter so it'd be a nice co-usage :)


keathley:
LGTM :tada:

Issue #169, opened 2017-03-25
Handling JS alert/confirm/prompt


padde:
For example, when generating a scaffold with Phoenix, we get a confirmation dialog via `data-confirm` for "Delete" links in the index view. If this link is clicked, Wallaby/PhantomJS will *not* perform the actual request, then usually the next assertion or selector fails.

For now, I use [this workaround](https://github.com/detro/ghostdriver/issues/20#issuecomment-13621554), which I put before the part that triggers the alert/confirm:

```elixir
# for alert()
execute_script(session, "window.alert = function(){}")

# for confirm()
execute_script(session, "window.confirm = function(){return true;}")
```

However, it seems clunky to use the above workaround, even if extracted into a helper to reduce duplication. So, my questions are:

* Is there a way to ignore/auto-accept dialog boxes globally?
* Any better way of handling this that I didn't think of?
* Is there a way to interact with PhantomJS directly, so that we can implement proper dialog handling before https://github.com/detro/ghostdriver/issues/20 is fixed?  (i.e. supply our own `onAlert` etc. handlers directly to Phantom)
* Should we add documentation on how to handle these situations?

If I can do anything to help with implementation or documentation, I'd be glad to.

padde:
So, after digging around a bit more I found that probably there is no nice way to handle this at the moment. The clean way would be to use the WebDriver interface to handle dialogs, but it seems that this won't happen soon.

The WebDriver spec [assumes that dialogs are blocking](https://www.w3.org/TR/webdriver/#user-prompts), so when a dialog appears the following command is expected to handle the dialog. If the subsequent command doesn't handle the dialog, an error is raised.

PhantomJS on the other hand assumes that dialogs are non-blocking. It expects a callback function such as [`page.onConfirm`](http://phantomjs.org/api/webpage/handler/on-confirm.html) before the actual dialog appears, where we can simulate user interaction with the dialog. This cannot trivially be made to block until the next command is received (accept/dismiss), therefore unlikely to be implemented soon in Ghostdriver. 

[...]

PR #173, opened 2017-03-28
Handle JavaScript dialogs

padde:
Implements JS dialog handling, as discussed in https://github.com/keathley/wallaby/issues/169.


test/support/pages/alert.html
```diff
@@ -1,21 +0,0 @@
...
-    function doPrompt() {
-      var name = prompt("name", "");
-        if (name != null) {
-            document.getElementById("result").innerHTML =
-            "Hello " + name;
-        }
-    }
-</script>
-  </body>
-</html>
```
padde: I felt free to rename and modify this fixture, since it wasn't actually used anywhere. Let me know if you're somehow attached to it or have other reasons to keep this as is 😉

padde:
Hey @keathley I've got an initial version done, would you mind taking a look?

keathley:
Thanks for this @padde! This looks pretty good to me. My only thought was that it might make sense to make `accept_alert` etc. chainable. I'm not sure if thats a good idea or not. Being able to see the message is definitely useful. I wonder if we could do that with another callback as the 3rd argument or if that api would just get messy. What do you think @PragTob?

lib/wallaby/browser.ex
```diff
@@ -980,6 +980,63 @@ defmodule Wallaby.Browser do
     current_url(session) == "about:blank"
   end
 
+  @doc """
+  Accepts all subsequent JavaScript dialogs in the given session.
+  """
```
PragTob: Thanks for adding docs right away! :+1:

lib/wallaby/browser.ex
```diff
@@ -980,6 +980,63 @@ defmodule Wallaby.Browser do
...
+  Dismisses all subsequent JavaScript dialogs in the given session.
+  """
+  def dismiss_dialogs(%Session{}=session) do
+    Driver.dismiss_dialogs(session)
+    session
+  end
+
+  @doc """
+  Accepts one alert dialog, which is triggered within the specified `fun`.
+  Returns the last message that was presented to the user.
```
PragTob: at least for the first one of those it might be nice to show a code example of what might be meant by `triggered by specified fun` :)

lib/wallaby/browser.ex
```diff
@@ -980,6 +980,63 @@ defmodule Wallaby.Browser do
...
+  Dismisses one confirm dialog, which is triggered within the specified `fun`.
+  Returns the last message that was presented to the user.
+  """
+  def dismiss_confirm(%Session{}=session, fun) do
+    Driver.dismiss_confirm(session, fun)
+  end
+
+  @doc """
+  Accepts one confirm dialog, which is triggered within the specified `fun`.
+  Returns the last message that was presented to the user.
```
PragTob: Should mention that this is a prompt an the input value needs to be specified :)

lib/wallaby/browser.ex
```diff
@@ -980,6 +980,63 @@ defmodule Wallaby.Browser do
...
+  """
+  def accept_prompt(%Session{}=session, options \\ [], fun) do
+    input_value = Keyword.get(options, :with)
+    Driver.accept_prompt(session, input_value, fun)
+  end
+
+  @doc """
+  Dismisses one confirm dialog, which is triggered within the specified `fun`.
```
PragTob: it's a prompt :)

test/support/pages/alert.html
PragTob: :+1:

test/wallaby/browser/dialog_test.exs
```diff
@@ -0,0 +1,208 @@
...
+    {:ok, %{page: page}}
+  end
+
+  describe "accept_dialogs/1" do
+    test "accept window.alert", %{page: page} do
+      result = page
+      |> accept_dialogs()
+      |> click(Query.link("Alert"))
+      |> find(Query.css("#result"))
+      |> Element.text()
```
PragTob: hm, this is mostly a style question and dunno how @keathley sees this but I prefer:

```
result = page
         |> thing
         |> other_thing
```

or 

```
result =
  page
  |> thing
  |> other_thing
```

Looks more tidy imo :)

PragTob:
Oh @keathley - having it chainable would be great, although with accept_dialogs being chainable I'm not sure how important it is as you can just use that if you don't care and then use the more detailed methods if you care about the values and then it's not so bad that it isn't chainable.

I'd like them to be chainable, but two functions as arguments seems weird as interface for a function. :|

padde:
Thanks for the comments guys, all valid points. Will push some improvements soon.

Regarding the chainable API issue let me summarize:

* `accept_alert(session, fun) -> string` (current solution)
  * not chainable
  * OTOH there is `accept_dialogs` for when we don't care about messages
* `accept_alert(session) -> session`, then `last_message(session) -> string`
  * easily implemented with PhantomJS (in fact the current solution internally does this)
  * not an option because it would be incompatible with the WebDriver spec, which only allows to get the message of the currently visible prompt *before* interacting with it
* `current_message(session) -> string`, then `accept_alert(session) -> session`
  * easily implemented with WebDriver
  * not an option because this is not possible with PhantomJS - it has no concept of "the current dialog". At the time when we know the message the alert is already accepted/dismissed
* `accept_alert(session, trigger_fun, message_callback_fun) -> session`
  * I agree having to supply two callback functions seems weird
  * I don't like having to trust the underlying code to actually invoke the assertion

On a more general note, I dislike the fact that the current solution mixes concerns by both interacting with the interface and querying it for the alert message. Alas, it seems near impossible to overcome the limitations of PhantomJS' implementation in this regard.

test/wallaby/browser/dialog_test.exs
padde: I actually prefer the style in the second example, but tried to stick to the way existing tests were written, because those that I had looked at were like this. As I read some more code and saw different styles being mixed, I'm not so sure which is the preferred style. Let @keathley decide 😉

lib/wallaby/browser.ex
padde: Duh, good catch :)

lib/wallaby/phantom/driver.ex
```diff
@@ -333,6 +333,135 @@ defmodule Wallaby.Phantom.Driver do
...
+    var page = this;
+    page.__onAlertDefault = page.onAlert;
+    page.onAlert = function(msg) {
+      page.__alertMessage = msg;
+      page.onAlert = page.__onAlertDefault;
+    }
+    return "ok";
+    """
+    {:ok, "ok"} = execute_phantom_script(session, script)
+    fun.(session)
```
padde: @keathley @PragTob what do you think about the session parameter being passed as argument to the callback function - do we really need this? Are there use cases where the session struct is changed through "normal usage" such that we need to make sure that we pass it along all the way?

test/wallaby/browser/dialog_test.exs
padde: I went with the the second version now, let me know if you had something else in mind @keathley.

padde:
@keathley @PragTob everything should be fixed now. Your opinion on [this comment](https://github.com/keathley/wallaby/pull/173#discussion_r108791947) might result in additional changes though.

padde:
Oh and maybe I should add a section to the README about handling JavaScript dialogs?

lib/wallaby/phantom/driver.ex
PragTob: At first I thought we didn't need it, as we have it in context most of the time, but then I thought of a case were you wanna implement as a separate method like

```
defp my_fun(session), do: #....
````

and then pass it as `&my_fun/1` - as it needs access to the session `&my_fun/0` wouldn't work so I like this :)

Comment (padde) [inline lib/wallaby/phantom/driver.ex]: Fair point. We could make the parameter optional, something along the lines of

```elixir
callback =
  case :erlang.fun_info(fun) |> Keyword.get(:arity) do
    0 -> fn (_) -> fun.() end
    1 -> fun
    _ -> raise BadArityError, args: [session], function: fun
  end
Driver.accept_alert(session, callback)
```

Not sure if this would be worth it though.

PragTob: yeah me neither, I think it's fairly common for people to just do `fn(_) -> i_dont_need_your_arguments end`

PragTob:
I think we could get this in now, thanks for the README section @padde - thoughts @keathley ? :)

padde:
Nice! Any ideas when we can expect a release?

By the way: awesome experience working with you guys, shows once again what a wonderful community this is. Greatly appreciated, thanks!

keathley:
@padde Thanks! I've added it to our 0.17.0 release milestone. Hopefully it won't take too long to get that out.

lib/wallaby/browser.ex
```diff
@@ -1040,6 +1040,105 @@ defmodule Wallaby.Browser do
...
+  ```
+  message = accept_prompt session, [with: "User input"], fn(s) ->
+    click(s, Query.link("Trigger prompt"))
+  end
+  ```
+  """
+  def accept_prompt(%Session{}=session, options \\ [], fun) do
+    input_value = Keyword.get(options, :with)
+    Driver.accept_prompt(session, input_value, fun)
+  end
```
padde: - [ ] Maybe we could add a safeguard here in case the supplied `input_value` is not a string, because  `window.prompt` will always return a string except when it's dismissed. For example, providing an integer as argument could be considered an error. Likewise, passing `nil` would be equivalent to calling `dismiss_prompt`, so that function should be used instead.

@PragTob @keathley what do you think? Should we raise an error here? Or maybe a warning would be more appropriate?

lib/wallaby/phantom/driver.ex
```diff
@@ -333,6 +333,135 @@ defmodule Wallaby.Phantom.Driver do
...
+    page.__onConfirmDefault = page.onConfirm;
+    page.onConfirm = function(msg) {
+      page.__confirmMessage = msg;
+      page.onConfirm = page.__onConfirmDefault;
+      return returnVal;
+    }
+    return "ok";
+    """
+    {:ok, "ok"} = execute_phantom_script(session, script, [return_value])
+    fun.(session)
```
padde: - [ ] When invoking the callbacks like here, do we need to catch errors to make sure that the callback is properly reset? IMO this would only be a problem if:

* inside the callback an error is raised *before* the dialog is triggered
* the test that actually uses this `accept_*`/`dismiss_*` function catches errors itself
* so when the error was catched in the surrounding test, the callback handler is in place, but was not reset back to the default handler

Might be an edge case that's not worth handling...

lib/wallaby/phantom/driver.ex
```diff
@@ -333,6 +333,135 @@ defmodule Wallaby.Phantom.Driver do
...
+    script = """
+    var page = this;
+    page.onConfirm = page.__onConfirmDefault;
+    return ["ok", page.__confirmMessage];
+    """
+    {:ok, ["ok", message]} = execute_phantom_script(session, script)
+    message
```
padde: - [ ] Instead of returning the message here we could still keep retrieval of messages as a separate function and then make the `accept_*`/`dismiss_*` functions chainable .

This would of course require us to store the messages somewhere, most likely in the `Session` struct. That would of course require the user to actually store the new state back in the `session`/`page` variable within the test, not sure if it's worth that. AFAICS this is not necessary with all of the other browser  functions so far.

For example the following would work:

```
session =
  session
  |> visit("dialogs.html")
  |> accept_alert(fn (s) ->
    click(s, Query.link("Trigger alert")
  end)

message = last_dialog_message(session)
```

but this would not work:

```
session
|> visit("dialogs.html")
|> accept_alert(fn (s) ->
  click(s, Query.link("Trigger alert")
end)

message = last_dialog_message(session)
```

lib/wallaby/phantom/driver.ex
```diff
@@ -333,6 +333,135 @@ defmodule Wallaby.Phantom.Driver do
...
+    return "ok";
+    """
+    {:ok, "ok"} = execute_phantom_script(session, script, [return_value, use_default])
+    fun.(session)
+    script = """
+    var page = this;
+    page.onPrompt = page.__onPromptDefault;
+    return ["ok", page.__promptMessage];
+    """
+    {:ok, ["ok", message]} = execute_phantom_script(session, script)
```
padde: By the way, this nested "fake :ok tuple" is necessary, because in case of a JavaScript error on the PhantomJS side, it will unfortunately give us a 200 OK response with the return value being a JS stack trace instead of the expected value. Maybe there's a better way to handle this somehow?

lib/wallaby/phantom/driver.ex
padde: Poltergeist for example tracks the occurred dialog messages within the JS client, might be worth looking into: https://github.com/teampoltergeist/poltergeist/blob/f7a8e970562fd1f912b39a0acc1afe7761d16a30/lib/capybara/poltergeist/client/browser.coffee#L73-L75

padde: Oh now I realize that I didn't like that approach because it would not work the same way with other drivers. We'd have to keep track of the messages for other browsers within Wallaby if we'd choose this API. Sorry for the confusion :/

lib/wallaby/browser.ex
PragTob: If it should always be a string, the easiest version (to my mind) would be to just have a `when` clause or something in `Driver.accept_prompt` that matches if it is a binary/string and otherwise either a match error or another implementation that raises ind is like "these are the arguments something went wrong"

lib/wallaby/phantom/driver.ex
PragTob: It's good that you think of that, and it might be nice if we ensured that it is reset correctly and then reraise the error or something. I think that'd be the most correct thing to do :)

However, we can do that in a separate PR or even only if the problem arises. E.g. it should only affect the current session (or so I hope/guess) and when an error is raised from that function it is already broken either way so the impact shouldn't be high.

lib/wallaby/phantom/driver.ex
PragTob: Keeping track of them in the session sounds like a great idea, much less clunky then a second function callback. they would all just go into a list the way they occured and could be accesses.

That people don't reassign the session shouldn't be a problem, it's immutable data structures and at some point people should know that and it's common to how session/conn work everywhere else as well.

lib/wallaby/phantom/driver.ex
PragTob: huh that seems unexpected. Not a huge fans of code comments most of the time, but this would be a great place for a WHY comment (like the one you wrote up just now) so we don't stumble upon it along the road :)

lib/wallaby/browser.ex
keathley: Is the string argument a requirement for the phantom driver? Or is that something that we should support at the Browser level? If its a part of the Browser api then we should have a guard in the Browser function. That way we can guard at the highest level possible.

lib/wallaby/phantom/driver.ex
keathley: I agree. Thats something that we can test for and fix as a separate PR. If it does throw an exception then I _think_ it should crash the test with the correct stack trace.

lib/wallaby/phantom/driver.ex
keathley: I actually like this approach much more as well. It allows us to chain the dialog functions without a clunky interface. We already do something similar with screenshots.

lib/wallaby/phantom/driver.ex
keathley: Do these stack traces get picked up by the JS Logger? Or are they contained? Should we provide a more meaningful `{:ok, value} | {:error, error}` response back to the `Browser` module so that we can provide a better error message for the end user?

lib/wallaby/browser.ex
padde: A `window.prompt` returns:

* `null` if dismissed
* `""` if accepted without input (and without a default value in place of course)

This means if we "accept" a prompt with a value of `null`, we actually dismiss it. There is no way to provide `null` or even non-string input to a prompt, so I'd say this should be handled on the Browser level.

lib/wallaby/phantom/driver.ex
padde: @keathley what do you think about the caveat outlined in my code examples? Could be tricky to debug if you write it like in the second example.

We don't have any place to store session state except in the `session` struct, right? This change makes it the User's responsibility to reassign `session` to the updated struct that contains the messages.

If that's OK with you, I could start working on this. Should happen before the next release though, otherwise we'd immediately have incompatible changes in the release after that one. I'd be OK with directly putting it in this PR here.

PragTob: Users have to reassign the session all the time, so I think it's fine as it's fairly common practice. Also they don't HAVE TO reassign, they can just pipe it into a function that then looks it up and returns the session again. I use that a lot when writing my tests.

So it's fine. If you prefer, to do it here you can but the PR is already fairly large so reviews get cumbersome so my personal preference is for an extra PR personally :)

lib/wallaby/phantom/driver.ex
padde: I'm not sure what you mean @keathley, but for clarification here's an example return value from `execute_phantom_script` when an error occurred:

```
IO.inspect execute_phantom_script(session, "boom")

{:ok,
 %{"line" => 1,
   "stack" => "anonymous\n_executePhantomJS@phantomjs://code/session.js:650:30\n_executePhantomJS@phantomjs://code/session_request_handler.js:863:68\n_handle@phantomjs://code/session_request_handler.js:166:30\n_reroute@phantomjs://code/request_handler.js:61:20\n_handle@phantomjs://code/router_request_handler.js:78:46"}}
```

The error message seems **not** to be included, just the line and stack trace. Any suggestions on how to improve this?

lib/wallaby/phantom/driver.ex
padde: I was thinking this PR for a single reason: getting the API right before merging. How about I make a separate PR that uses this PR as base. You can review it in isolation and when it's good it can be merged into this PR. Once we're all set to release we can merge this PR with all the changes related to dialogs.

lib/wallaby/phantom/driver.ex
keathley: So typically if there are javascript errors in the page those will get picked up by wallaby's internal JS logger. If you execute a script does it get picked up by that logger? I'm guessing that it doesn't.

padde: How can I best access the output JS Logger? If we use the "normal" `execute_script` function to trigger an error, it will not crash. But it returns something like this, which at least includes the original error message.


```
IO.inspect execute_script(session, "boom")

%{"class" => "org.openqa.selenium.WebDriverException",
  "message" => "{\"errorMessage\":\"Can't find variable: boom\",\"request\":{\"headers\":{\"Content-Length\":\"27\",\"Content-Type\":\"text/json, text/json\",\"Host\":\"localhost:51992\",\"User-Agent\":\"hackney/1.6.6\"},\"httpVersion\":\"1.1\",\"method\":\"POST\",\"post\":\"{\\\"script\\\":\\\"boom\\\",\\\"args\\\":[]}\",\"url\":\"/execute\",\"urlParsed\":{\"anchor\":\"\",\"query\":\"\",\"file\":\"execute\",\"directory\":\"/\",\"path\":\"/execute\",\"relative\":\"/execute\",\"port\":\"\",\"host\":\"\",\"password\":\"\",\"user\":\"\",\"userInfo\":\"\",\"authority\":\"\",\"protocol\":\"\",\"source\":\"/execute\",\"queryKey\":{},\"chunks\":[\"execute\"]},\"urlOriginal\":\"/session/5cb946a0-19c4-11e7-aef6-3973bb41a47b/execute\"}}",
  "screen" => "<Large Base64 encoded string ...>"}
```


Since both `execute_script` and `execute_phantom_script` internally use `check_logs!` I suppose that is what you mean with the internal JS Logger. AFAICS both these return values already come from the logger.

It would be really nice to raise an error in Wallaby if an exception is raised in  `execute_script` or `execute_phantom_script`.

lib/wallaby/phantom/driver.ex
PragTob: it's fine if you prefer this PR, get the basic implementation in a separate commit then it's also rather easy to review. Thanks a lot :)

padde: @PragTob @keathley I am not so sure any more that this is worth it. The initial idea was that we wouldn't have to "interrupt" the pipeline in order to get the message later, but this approach kind of fails to live up to the expectations. I was trying out the approach just now and it turns out that instead of writing the test case like this:

```elixir
test "accept window.alert and get message", %{page: page} do
  message = accept_alert page, fn(p) ->
    click(p, Query.link("Alert"))
  end
  result =
    page
    |> find(Query.css("#result"))
    |> Element.text()
  assert message == "This is an alert!"
  assert result == "Alert accepted"
end
```

I now have to write it like this, because the modified `session` needs to be stored before we can continue:

```elixir
test "accept window.alert and get message", %{page: page} do
  page =
    page
    |> accept_alert(fn(p) ->
      click(p, Query.link("Alert"))
    end)
  result =
    page
    |> find(Query.css("#result"))
    |> Element.text()
  assert last_dialog_message(page) == "This is an alert!"
  assert result == "Alert accepted"
end
```

I am honestly wondering if this is any better than before, in particular if this is worth the additional complexity, having to keep track of messages in the session struct. Basically, this only simplifies the case you're using the finer grained `accept_alert` etc functions, but are not interested in the message.

It would be great to have some feedback from you before I actually dig in and implement this.

lib/wallaby/phantom/driver.ex
keathley: Yes I agree. We can open a new issue for that.

lib/wallaby/phantom/driver.ex
keathley: Don't worry about implementing this for now. Lets just return the alert text from `accept_alert` and the other helpers. Once the api has seen some serious use then we'll know better what direction we want to go.

padde: The pragmatic approach 😉 I'm fine leaving this as it is, thanks for the feedback guys.

keathley:
@padde I'm going to do a final review of everything here, but is there anything else that you think we need to look at before we merge?

padde:
@keathley I'd like to add the guard for the `input_value` in `accept_prompt` (see https://github.com/keathley/wallaby/pull/173#discussion_r109278665) before merging. Other than that we should be good to go.

padde:
@keathley added the guard for the `input_value` with `accept_prompt/3`, should be good to go now I think.

PragTob:
Okay I think this should be good to go now, besides - we might need it at work :D 

I'll try to dig through and file some issues we've mentioned.

PragTob:
Ok just found one case where I really want to open an issue, if I missed any others please feel free tom open them :)

PR #175, opened 2017-03-30
Reword AcceptanceCase to FeatureCase

### LINKED ISSUE #167: Is `AcceptanceCase` really the right term?
According to the [Wikipedia article on software testing](https://en.wikipedia.org/wiki/Software_testing#Acceptance_testing),

> Acceptance testing can mean one of two things:
>
> 1. A smoke test is used as an acceptance test prior to introducing a new build to the main testing process, i.e., before integration or regression.
> 2. Acceptance testing performed by the customer, often in their lab environment on their own hardware, is known as user acceptance testing (UAT). Acceptance testing may be performed as part of the hand-off process between any two phases of development.

This leads me to the conclusion that `AcceptanceCase` is not really the right term for what we're doing with Wallaby.

Maybe one the following alternatives would be a better fit:

* `IntegrationCase`
* `EndToEndCase`
* `BrowserCase`
* `FunctionalCase`
* `SystemCase`

`IntegrationCase` sounds better IMO, but integration testing also refers to testing just parts of a system in conjunction, as opposed to unit testing (just one function) or system testing (the entire thing).

Both `EndToEndCase` and `BrowserCase` would refer to the fact that we are testing through the browser, exercising the entire stack end-to-end.

`FunctionalCase` is not really my favorite, since the term "functional test" seems to be used on a different level than the others. It's more a question of "do we test user requirements directly" (functional) or "do we test some part of the system that is not really important from a user's perspective". On top of that, the term "functional" might be too confusing. Nevertheless, I've sometimes seen "functional test" used as a term for browser tests, so I dropped it in here.

Finally, `SystemCase` would mean that we are testing the entire system, not just parts thereof.

Discuss! 😉  What are your thoughts?

padde:
Fixes #167

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,7 +89,7 @@ config :wallaby, phantomjs_args: "--webdriver-logfile=phantomjs.log"
 Its easiest to add Wallaby to your test suite by creating a new Case Template:
 
 ```elixir
-defmodule YourApp.AcceptanceCase do
+defmodule YourApp.FeatureCase do
   use ExUnit.CaseTemplate
 
   using do
@@ -123,7 +123,7 @@ Then you can write tests like so:
 
 ```elixir
 defmodule YourApp.UserListTest do
-  use YourApp.AcceptanceCase, async: true
+  use YourApp.FeatureCase, async: true
 
   test "users have names", %{session: session} do
     first_employee =


keathley:
LGTM :tada:

Issue #177, opened 2017-03-31
Remove all deprecated functions


keathley:
We need to update Wallaby's internal test suite and remove the deprecated functions.

padde:
Hey @keathley I saw all the deprecations while working on the JS dialogs. I already thought about making a PR to rewrite the tests with the new API. Could you use some help here?

What about tests that test the deprecated functions themselves, should they stay or be removed altogether?

PragTob:
What we already wanna remove them all in 0.17? That's cool, wasn't sure what your deprecation policy would be :D 

I'd be super glad to remove them. Lots of them are not covered and get the test coverage down but of course writing tests for them doesn't make much sense :)

keathley:
@padde Any help here would be great! Some of the old tests should be updated to use the new api. But a lot of them can just be deleted. For instance, all the tests like, "fill_in/3 finds elements by id" and "fill_in/3 finds elements by label text" can be deleted.

At some point I'm going to wrap the entire external api with property tests so we don't have to write all of these permutations by hand any more.

padde:
@keathley sounds great, I'd be glad to help. Just to be clear: do you want to keep the deprecated methods themselves in `Browser` or should they be removed as well? If they're kept, should there be at least one test for each deprecated method left in the test suite?

For the deprecation policy, SemVer might give some guidance here (if you should decide to stick with this). It generally forbids backwards incompatible changes in minor releases, but:

> Major version zero (0.y.z) is for initial development. Anything may change at any time. The public API should not be considered stable.

keathley:
If its easier you can just remove it all. But if you just want to work on tests that fine.

It was an interesting decision to deprecate things before we hit 1.0 but I knew that there were enough people using it and the api was a big enough change that I wanted to provide the deprecations as a way to help people migrate. Now that we have a good direction for the api I'm fine with removing the old functionality. Plus it gets us closer to a 1.0 release 😉

PragTob:
Very pragmatic decision, I might get onto removing some of them during the weekend (I'd try to do it in smaller chunks)

padde:
Hey @PragTob didn't want to steal this from you, please go ahead! I think it makes more sense to  remove deprecated functions along with their tests and also replace them in other tests in one go. The compiler will basically tell you exactly what to do once you've removed the function, and we'd probably trip over each others feet all the time when merging.

PragTob:
[...]

PR #181, opened 2017-04-01
Remove deprecated functions - fill_in

PragTob:
A first step toward #177 

As this is the first one I thought I'd keep it small and simple (I just started with the first function that had an `IO.warn` in there) so we can see if this is a good template for removing outdated functions :)

### DIFF
--- a/lib/wallaby/browser.ex
+++ b/lib/wallaby/browser.ex
@@ -155,29 +155,12 @@ defmodule Wallaby.Browser do
   end
 
   @doc """
-  Fills in a "fillable" element with text. Input elements are looked up by id, label text,
-  or name.
+  Fills in an element identified by `query` with `value`.
   """
-  @spec fill_in(element, opts) :: element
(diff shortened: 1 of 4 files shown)


lib/wallaby/browser.ex
```diff
@@ -155,30 +155,22 @@ defmodule Wallaby.Browser do
...
-    |> find(Query.fillable_field(locator, opts), &(Element.fill_in(&1, with: value)))
-  end
+  ### Examples
+
+      page
+      |> fill_in(Query.text_field("name"), with: "Chris")
+      |> fill_in(Query.css("#password_field", with: "secret42"))
+
+  """
+  @spec fill_in(parent, Query.t, with: String.t) :: parent
```
keathley: This makes me so happy.

keathley:
This looks good to me. It seems like it should be straightforward to make incremental progress.

PragTob:
Then I'll just merge and whoever/whenever can take a stab at the next function(s). I'd say we sync in the issue. :)

PR #183, opened 2017-04-03
make execute_script return session instead of script result

### LINKED ISSUE #172: Docs/spec confusing for functions with return type parent
Here's an example of what I was trying to do:
```
session
|> visit(foo_path(Endpoint, :new))
|> execute_script("...")
|> take_screenshot()
```
This raises `(UndefinedFunctionError) function nil.url/0 is undefined or private` and wondering what's that about. I had to look at the source to figure out that `execute_script` returns the value and not the session even though the spec `execute_script(parent, String.t, list) :: parent` makes me think it would return the session.

Side note, it would be nice to have functions like `execute_script` be pipe-able or a pipe-able version  that ignores the return value like my example. For now I'm doing:
```
def execute_script_pipe(session, script, arguments \\ []) do
   {:ok, _} = Driver.execute_script(session, script, arguments)
   session
end
```

krankin:
This resolves #172 by changing ```execute_script``` to return the session passed to it and allowing a callback to be passed to ```execute_script```  which has the script's return value as its parameter.

### DIFF
--- a/lib/wallaby/browser.ex
+++ b/lib/wallaby/browser.ex
@@ -527,13 +527,30 @@ defmodule Wallaby.Browser do
 
   @doc """
   Executes javascript synchoronously, taking as arguments the script to execute,
-  and optionally a list of arguments available in the script via `arguments`
+  an optional list of arguments available in the script via `arguments`, and an
+  optional callback function with the result of script execution as a parameter.
   """
+  @spec execute_script(parent, String.t) :: parent
   @spec execute_script(parent, String.t, list) :: parent
+  @spec execute_script(parent, String.t, ((binary()) -> any())) :: parent
+  @spec execute_script(parent, String.t, list, ((binary()) -> any())) :: parent
 
-  def execute_script(session, script, arguments \\ []) do
+
+  def execute_script(session, script) do
+    execute_script(session, script, [])
+  end
+
+  def execute_script(session, script, arguments) when is_list(arguments) do
+    execute_script(session, script, arguments, fn(value) -> end )
+  end
+  def execute_script(session, script, callback) when is_function(callback) do
+    execute_script(session, script, [], callback)
+  end
+
+  def execute_script(session, script, arguments, callback) when is_list(arguments) and is_function(callback) do
     {:ok, value} = Driver.execute_script(session, script, arguments)
-    value
+    callback.(value)
+    session
   end
 
   @doc """
--- a/test/wallaby/browser/local_storage_test.exs
+++ b/test/wallaby/browser/local_storage_test.exs
@@ -16,24 +16,32 @@ defmodule Wallaby.Browser.LocalStorageTest do
     |> visit("index.html")
     |> execute_script(@set_value_script)
 
-    assert session
-    |> execute_script(@get_value_script) == "foo"
+    session
+    |> execute_script(@get_value_script, fn(value) -> send self(), {:result, value} end)
 
-    assert s2
-    |> visit("index.html")
-    |> execute_script(@get_value_script) == nil
+    assert_received {:result, "foo"}
 
-    assert s3
+    s2
     |> visit("index.html")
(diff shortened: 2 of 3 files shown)


lib/wallaby/browser.ex
```diff
@@ -527,13 +527,35 @@ defmodule Wallaby.Browser do
...
+  def execute_script(session, script, arguments) when is_list(arguments) do
+    {:ok, _} = Driver.execute_script(session, script, arguments)
+    session
+  end
+
+  def execute_script(session, script, callback) when is_function(callback) do
+    {:ok, value} = Driver.execute_script(session, script, [])
+    callback.(value)
+    session
+  end
```
PragTob: Hm not sure, this seems like we are implementing the same function 4 times (3 here + the one below). I'm wondering if we could do anything through the use of default arguments. E.g. arguments defaults to `[]` and function defaults to `nil`.

I'd be fine with the callback function only being available if you pass in the arguments explicitly (even if it's `[]`). That should work then.

Alternatively the more abstracted functions (without args or callback functions) could call the more detailed functions to avoid the duplication.

Wdyt @keathley ? :)

test/wallaby/browser/local_storage_test.exs
```diff
@@ -16,24 +16,32 @@ defmodule Wallaby.Browser.LocalStorageTest do
     |> visit("index.html")
     |> execute_script(@set_value_script)
 
-    assert session
-    |> execute_script(@get_value_script) == "foo"
+    session
+    |> execute_script(@get_value_script, fn(value) -> send self(), {:result, value} end)
+
+    assert_received {:result, "foo"}
```
PragTob: any reason against using the assert straight inside the callback function to avid the send work around?

krankin: I couldn't guarantee that the assert was being called in the callback. If I removed the code for calling the callback in execute_script, the assert isn't checked
``` Elixir
  def execute_script(session, script, callback) when is_function(callback) do
    {:ok, value} = Driver.execute_script(session, script, [])
    #callback.(value)  -- don't execute the callback
    session


-- In test: 

  result =
      session
      |> visit("page_1.html")
      |> execute_script(@script, ["now you see me", "return value"], fn(value) ->
           assert value == "return value"  #-- this still passes, is never actually asserted
         end)
  end
```

lib/wallaby/browser.ex
krankin: I agree that something like this looks much cleaner

```Elixir
  def execute_script(session, script) do
    execute_script(session, script, [])
  end

  def execute_script(session, script, arguments) when is_list(arguments) do
    execute_script(session, script, arguments, fn(value) -> end )
  end

  def execute_script(session, script, callback) when is_function(callback) do
    execute_script(session, script, [], callback)
  end

  def execute_script(session, script, arguments, callback) when is_list(arguments) and is_function(callback) do
    {:ok, value} = Driver.execute_script(session, script, arguments)
    callback.(value)
    session
  end
```

@keathley -- what are your thoughts about the callback only being available if you pass in all arguments?

test/wallaby/browser/local_storage_test.exs
PragTob: good point, thanks for clarifying! :)

lib/wallaby/browser.ex
keathley: @krankin I like your example implementation. Thats a pattern we've used elsewhere in the api. The only drawback is that I think the typespecs will have to be something like `@spec execute_script(parent(), binary(), any()) :: parent()` or we'll need to create an opaque type like: `@opaque args_or_cb :: list() | ((any()) -> any())`. But I'm OK with that if you are @PragTob.

PragTob: Fine by me, I'd just like to group the same functions (same arity) together (no newline inbetween)

krankin:
changes made per conversation.

PR #184, opened 2017-04-03
Fix dialyzer errors and add dialyzer to build

aaronrenner:
These changes fix a few dialyzer errors in the project and also run dialyzer during the Travis CI build. I made a couple notable changes below

I removed the typespec from the `assert_has` and `refute_has` macros, because dialyzer was raising a warning on them. I checked in the Elixir, Ecto, Plug and Phoenix libraries and couldn't find a single example of where someone documented macros using a typespec.

I also changed the call to `Port.open/2` so it passes a charlist as the second element in the `{:spawn_executable, _}` tuple. This call passed directly to the `open_port/2` erlang function and expects an erlang string.

Please let me know what you think and if there's any changes you'd like me to make.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,17 @@
 language: elixir
 elixir: 1.4.0
 otp_release: 19.2
+cache:
+  directories:
+    - _build
+    - deps
+before_script:
+  - MIX_ENV=test mix compile --warnings-as-errors
+  - mix dialyzer --plt
 
 script:
   - mix coveralls.travis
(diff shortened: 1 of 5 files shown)


aaronrenner:
It looks like travis timed out for the elixir 1.3.2 build and I don't have permissions to retry the build.

PragTob:
:heart: to have dialyzer in the build! Thank you very much! :)

Too bad about the macros... there ought to be a way but that's rather something to take up with the elixir forum together with the appropriate error message :)

PR #187, opened 2017-04-04
Fix dialyzer on travis ci

### LINKED ISSUE #186: Reinstate dialyzer spec checking in travis build
Removed in #185 due to random failures would be great to have again.

aaronrenner:
This runs the build with `sudo: required` so it runs on a larger vm with [7.5GB of memory](https://docs.travis-ci.com/user/ci-environment/#Virtualization-environments), so mix dialyzer --plt doesn't run the vm out of memory and exit with code 137.

This also runs wraps the `mix dialyzer --plt` in the `travis_wait` command so if it doesn't output any log statements in 10 minutes, the build won't timeout.

Finally, the build is already being cached, so the only time the PLT needs to be built is on a fresh cache, or if the build matrix is changed.

Fixes #186.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,15 +1,19 @@
 language: elixir
 elixir: 1.4.0
 otp_release: 19.2
+# Run in bigger container so we don't run out of memory generating the plt
+sudo: required
 cache:
   directories:
     - _build
     - deps
 before_script:
   - MIX_ENV=test mix compile --warnings-as-errors
+  - travis_wait mix dialyzer --plt
 
 script:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,15 +1,19 @@
 language: elixir
 elixir: 1.4.0
 otp_release: 19.2
+# Run in bigger container so we don't run out of memory generating the plt
```
PragTob: :+1: on the WHY comment

PragTob:
Thanks a ton, of course! :tada: :+1:

PR #189, opened 2017-04-04
Update elixir versions in .travis.yml

PragTob:
Being up to date with the good ol patch levels.

Which reminds me, is there a reason we don't just run both elixir versions against both otp releases? Being very nice to travis? :) @keathley

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 language: elixir
-elixir: 1.4.0
+elixir: 1.4.2
 otp_release: 19.2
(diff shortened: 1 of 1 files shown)


keathley:
LGTM :tada:

@PragTob Someone else built the travis matrix for me. Thats my excuse ;)

aaronrenner:
@PragTob It might be worth updating .tool-versions, so asdf-based development environments update to elixir 1.4.2 as well.

PragTob:
@aaronrenner ah forgot about that (was on the web interface) - will do, happy asdf user myself :tada:

PR #190, opened 2017-04-04
Updated hackney and httpoison in mix.lock

aaronrenner:
The hackney dependency is locked to 1.6.6 which has been retired. Instead of keeping httpoison at the same version same and downgrading to hackney 1.6.5, I figured we might as well update to httpoison 0.11.1, which upgrades hackney to 1.7.1.


PragTob:
:+1:

PR #191, opened 2017-04-05
Added tests for Wallaby.Phantom.Driver

aaronrenner:
This change tests the interactions between the Wallaby.Phantom.Driver module and the PhantomJS webdriver API. I wanted to make sure I had test coverage here so I didn't accidentally break the serialization/deserialization logic when I start extracting logic to add a Selenium driver.

Another couple of changes to note:

  * Changed the Content-Type header from "text/json" to "application/json". See: http://stackoverflow.com/a/477819
  * Extracted a create_session/2 function in Wallaby.Phantom.Driver that only handles the API interaction. This made it much easier to test with Bypass. I also used @doc false because I didn't want to expose a new public function in the docs.


keathley:
Thanks @aaronrenner! It'll be nice to have some tests around just the driver.

LGTM :tada:

test/wallaby/phantom/driver_test.exs
```diff
@@ -0,0 +1,736 @@
...
+
+          send_resp(conn, 200, ~s<{
+            "sessionId": "#{session.id}",
+            "status": 0,
+            "value": null
+          }>)
+        end
+      end
+
+      # Returns only the value
```
keathley: I'm not sure what these comments are really telling me. Do we need them still?

lib/wallaby/phantom/driver.ex
```diff
@@ -14,6 +14,8 @@ defmodule Wallaby.Phantom.Driver do
   @type params :: %{using: String.t, value: query}
   @type locator :: Session.t | Element.t
 
+  # Start a Session for the given Phantom.Server
```
keathley: Can we remove this comment or move it into a `@doc` string?

test/wallaby/phantom/driver_test.exs
aaronrenner: Just removed them. 🙂

PR #194, opened 2017-04-06
Test to ensure StaleReferenceException is raised

aaronrenner:
I was curious what the response looked like from PhantomJS when we get a StaleReferenceException found out we didn't have a test triggering it.


keathley:
If the sleep does cause the test to become flaky then I think we can fix it by using `refute_has(session, Query.css("#removed-node"))` to force the test to block until the node is removed from the DOM.

PR #195, opened 2017-04-07
Extracted Wallaby.Webdriver.Client module

aaronrenner:
This isolates all of our HTTP interactions, and serializations/deserializations into a single module. This will make it easier to test and globally parse errors that come back from the webdriver API. I left the http tests on Wallaby.Phantom.Driver for the time being to make sure as I do further refactorings, I don't break anything.

### DIFF
--- a/lib/wallaby/phantom/driver.ex
+++ b/lib/wallaby/phantom/driver.ex
@@ -7,12 +7,8 @@ defmodule Wallaby.Phantom.Driver do
   alias Wallaby.Element
   alias Wallaby.Phantom.Logger
   alias Wallaby.Phantom.LogStore
-
-  @type method :: :post | :get | :delete
-  @type url :: String.t
-  @type query :: String.t
-  @type params :: %{using: String.t, value: query}
-  @type locator :: Session.t | Element.t
+  alias Wallaby.Webdriver.Client
+  alias Wallaby.Phantom. Webdriver.Client, as: PhantomClient
 
   @spec create(pid, Keyword.t) :: {:ok, Session.t}
   def create(server, opts) do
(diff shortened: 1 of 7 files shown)


keathley:
Thanks @aaronrenner. Originally I was thinking that we could extract just the request interactions. Since each driver has slightly different implementations this would allow us to make whichever requests were needed in each driver. I'm worried that we'll run into issues with discrepancies between drivers if they all try to use the same webdriver client. For instance, some of the endpoints that we use in the phantom driver are phantom specific and IIRC won't work with selenium. What are your thoughts?

aaronrenner:
@keathley I was trying to extract all of the WebDriver API serialization/request/deserialization code into Wallaby.Webdriver.Client, but leave the specific logic for things like `check_logs!` in the specific driver implementation. This way we could share common parsing logic between the PhantomDriver and SeleniumDriver implementations and when we get an `"org.openqa.selenium.StaleElementReferenceException"`it will be returned the same way across drivers. Do you see us running into an issue where the same endpoint would need to be parsed one way for Phantom and another way for Selenium?

Maybe a better approach would be for me to build out the SeleniumDriver implementation without sharing code and then extract the common code into a module.

aaronrenner:
I just rebased on my changes on top of the new changes to handle phantomjs alerts. Since this new code uses a phantom specific endpoint, I moved that code to `lib/wallaby/phantom/webdriver/client.ex`. And again the nice thing about the main http interaction being extracted is I don't have to mock the `/log` endpoint to the `check_logs!` function will run. 

Please let me know if you'd like me to structure or name things differently.

aaronrenner:
That makes sense. I'll build the separate modules for each driver first and then extract common functionality afterwords.

PR #197, opened 2017-04-14
Remove `choose/n` in favor of `click/2` + `Query.radio_button`

PragTob:
Put the tests into the `click_test` for now in a separate describe, reproduced most of them but not all of them as some of it was fairly common place functionality that doesn't need extra tests :)


keathley:
LGTM :tada:

Issue #95, opened 2016-09-07
Capture JS errors and logs in the servers `handle_info` call


keathley:
Blocking each webdriver call with a call to check for js errors and logs is an expensive and wasteful process. All of this same information is sent to the server process from phantom. We should be able to capture this information using the same logging capture behaviour and then print logs to the console or throw errors in our respective test. This **should** work correctly because each phantom is scoped to a single test at a time.

krankin:
@keathley - So I was playing around with this and had a question about how you want to handle javascript errors on the Genserver side?  In the case where we have a javascript error, raising a ```Wallaby.JSError``` goes uncaught and crashes the Genserver if you move current logging behavior server side.  

We could do something to notify the client, but doesn't that kinda defeat the purpose of this as we'd still be checking for an error on each call in the driver? I still think we can handle logging server side,  but I'm looking to you for ideas on what to change for handling js errors?

PR #198, opened 2017-04-14
Experimental selenium driver

aaronrenner:
This is a work in progress to add selenium support and allows the existing test suite to be run against a local instance of geckodriver or chromedriver. This is definitely a work in progress, but will show us how the existing test suite runs against different webdriver backends.

### DIFF
--- a/integration_test/cases/browser/click_button_test.exs
+++ b/integration_test/cases/browser/click_button_test.exs
@@ -1,6 +1,8 @@
 defmodule Wallaby.Integration.Browser.Actions.ClickButtonTest do
   use Wallaby.Integration.SessionCase, async: true
 
+  alias Wallaby.Integration.PageObjects.IndexPage
+
   setup %{session: session} do
     page =
       session
@@ -13,54 +15,60 @@ defmodule Wallaby.Integration.Browser.Actions.ClickButtonTest do
     current_url =
       page
       |> click_button("Submit button")
+      |> IndexPage.ensure_page_loaded
       |> current_url
 
-    assert current_url == "http://localhost:#{URI.parse(current_url).port}/index.html"
+    assert current_url =~ "http://localhost:#{URI.parse(current_url).port}/index.html"
   end
 
   test "clicking button with no type via name (submits form)", %{page: page} do
     current_url =
       page
       |> click_button("button-no-type")
+      |> IndexPage.ensure_page_loaded
       |> current_url
 
-    assert current_url == "http://localhost:#{URI.parse(current_url).port}/index.html"
+    assert current_url =~ "http://localhost:#{URI.parse(current_url).port}/index.html"
   end
 
   test "clicking button with no type via id (submits form)", %{page: page} do
     current_url =
       page
       |> click_button("button-no-type-id")
+      |> IndexPage.ensure_page_loaded
       |> current_url
 
-    assert current_url == "http://localhost:#{URI.parse(current_url).port}/index.html"
+    assert current_url =~ "http://localhost:#{URI.parse(current_url).port}/index.html"
   end
 
   test "clicking button type[submit] via button text (submits form)", %{page: page} do
     current_url =
       page
       |> click_button("Submit button")
+      |> IndexPage.ensure_page_loaded
       |> current_url
 
-    assert current_url == "http://localhost:#{URI.parse(current_url).port}/index.html"
+    assert current_url =~ "http://localhost:#{URI.parse(current_url).port}/index.html"
   end
 
   test "clicking button type[submit] via name (submits form)", %{page: page} do
     current_url =
       page
       |> click_button("button-submit")
+      |> IndexPage.ensure_page_loaded
       |> current_url
 
-    assert current_url == "http://localhost:#{URI.parse(current_url).port}/index.html"
+    assert current_url =~ "http://localhost:#{URI.parse(current_url).port}/index.html"
   end
 
   test "clicking button type[submit] via id (submits form)", %{page: page} do
     current_url =
       page
       |> click_button("button-submit-id")
+      |> IndexPage.ensure_page_loaded
       |> current_url
 
-    assert current_url == "http://localhost:#{URI.parse(current_url).port}/index.html"
+    assert current_url =~ "http://localhost:#{URI.parse(current_url).port}/index.html"
   end
 
   test "clicking button type[button] via button text (resets input via JS)", %{page: page} do
@@ -133,27 +141,30 @@ defmodule Wallaby.Integration.Browser.Actions.ClickButtonTest do
     current_url =
       page
       |> click_button("Submit input")
+      |> IndexPage.ensure_page_loaded
       |> current_url
 
-    assert current_url == "http://localhost:#{URI.parse(current_url).port}/index.html"
+    assert current_url =~ "http://localhost:#{URI.parse(current_url).port}/index.html"
   end
 
   test "clicking input type[submit] via name submits form", %{page: page} do
     current_url =
       page
       |> click_button("input-submit")
+      |> IndexPage.ensure_page_loaded
       |> current_url
 
-    assert current_url == "http://localhost:#{URI.parse(current_url).port}/index.html"
+    assert current_url =~ "http://localhost:#{URI.parse(current_url).port}/index.html"
   end
 
   test "clicking input type[submit] via id submits form", %{page: page} do
     current_url =
       page
       |> click_button("input-submit-id")
+      |> IndexPage.ensure_page_loaded
       |> current_url
 
-    assert current_url == "http://localhost:#{URI.parse(current_url).port}/index.html"
+    assert current_url =~ "http://localhost:#{URI.parse(current_url).port}/index.html"
   end
 
   test "clicking input type[button] via button text resets input via JS", %{page: page} do
@@ -226,18 +237,20 @@ defmodule Wallaby.Integration.Browser.Actions.ClickButtonTest do
     current_url =
       page
       |> click_button("input-image")
+      |> IndexPage.ensure_page_loaded
       |> current_url
 
-    assert current_url == "http://localhost:#{URI.parse(current_url).port}/index.html"
+    assert current_url =~ "http://localhost:#{URI.parse(current_url).port}/index.html"
   end
 
   test "clicking input type[image] via id submits form", %{page: page} do
     current_url =
       page
       |> click_button("input-image-id")
+      |> IndexPage.ensure_page_loaded
       |> current_url
 
-    assert current_url == "http://localhost:#{URI.parse(current_url).port}/index.html"
+    assert current_url =~ "http://localhost:#{URI.parse(current_url).port}/index.html"
   end
 
   test "waits until the button appears", %{page: page} do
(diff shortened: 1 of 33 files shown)


lib/wallaby/experimental/selenium.ex
```diff
@@ -0,0 +1,56 @@
...
+    {:remote_url, String.t} |
+    {:capabilities, map} |
+    {:create_session_fn, ((String.t, map) -> {:ok, %{}})}
+
+  @spec start_session([start_session_opts]) :: {:ok, Session.t}
+  def start_session(opts \\ []) do
+    base_url = Keyword.get(opts, :remote_url, "http://localhost:4444/wd/hub/")
+    capabilities = Keyword.get(opts, :capabilities, %{})
+    create_session_fn = Keyword.get(opts, :create_session_fn,
+                                    &Wallaby.Phantom.Driver.create_session/2)
```
aaronrenner: This is still calling `Wallaby.Phantom.Driver.create_session` to make the http request to selenium to start the session. Do you think all of the webdriver http requests should live in a common module like in #195, or would you prefer if each driver implements its own http requests?

aaronrenner:
I'd also be interested in how you'd like to see the driver modules designed. Should I be writing each one of the browser interaction functions (like `visit`, `click`, etc) in each module to make them easier to customize behavior per driver? Since chromedriver is different than geckodriver and phantomjs (and Appium for that matter), should I have one driver per backend? If so, do you think each module should do something like `use Wallaby.WebDriver.Driver` to share common boilerplate and then override functions when we need to change something? I'm not sure if using `use Wallaby.WebDriver.Driver` to share boilerplate functionality would make things to confusing.

I'm looking forward to hearing your ideas so I can build it right the first time.

lib/wallaby/experimental/selenium.ex
keathley: Ideally they could live in their own module but for now lets separate them. Once we have support for 3 drivers (phantomjs, selenium, chromedriver, etc.) then we can see if they share enough functionality to merge the driver implementations.

lib/wallaby/session.ex
```diff
@@ -44,11 +44,12 @@ defmodule Wallaby.Session do
     id: String.t,
     session_url: String.t,
     url: String.t,
-    server: pid(),
-    screenshots: list
+    server: pid | nil,
+    screenshots: list,
+    driver: module
```
keathley: I'm a big fan of this idea. This gives us a lot of flexibility for running different browsers against each other.

test/support/session_case.ex
```diff
@@ -8,7 +8,19 @@ defmodule Wallaby.SessionCase do
   end
 
   setup do
-    {:ok, session} = Wallaby.start_session
+    session_opts = case System.get_env("WEBDRIVER") do
```
keathley: I realize that this is just for testing so it doesn't matter right now, but if we decide to stick with an implementation like this we should probably prefix the env variable with `WALLABY_`. Totally not important right now though.

keathley:
I think for now we shouldn't try to share functionality between the different driver implementations. Once we have multiple implementations we can see exactly where the common code / patterns are and we can use shared code at that point. Its easier to remove common code once you can see all of it.

For now it *would* be a good idea to share a common api between the drivers. We should create a behaviour that each driver needs to conform to. If you want I can work on that tonight. Then we can use that as a basis for the other drivers.

test/support/session_case.ex
aaronrenner: Just changed the variable name to `WALLABY_WEBDRIVER`.

aaronrenner:
I was thinking all of the drivers would share the following functions:

```
create_session/2
delete_session/1
find_elements/2
set_value/2
clear/1
click/1
text/1
page_title/1
attribute/1
visit/2
current_url/1
current_url!/1
current_path!1/1
selected/1
displayed/1
displayed!/1
size/1
rect/1
take_screenshot/1
cookies/1
set_cookies/3
set_windows_size/3
get_windows_size/1
execute_script/3
send_keys/2
log/1
accept_dialogs/1
dismiss_dialogs/1
accept_alert/2
accept_confirm/2
dismiss_confirm/2
accept_prompt/3
dismiss_prompt/2
```

If a driver didn't support one of these functions, we could return `{:error, :unsupported_command}` or raise an exception. Is that what you were thinking?

keathley:
That looks about right. I think we can get rid of the `!` functions. Once #95 lands (which I think @krankin is working on) we can remove `logs` from that list as well.

aaronrenner:
I just checked in the `Wallaby.Driver` behavior and switched `Wallaby.Element` and `Wallaby.Browser` to get the current driver off the session in #201. I think this will make adding a Selenium driver much easier.

aaronrenner:
I finally got the tests passing and this is code should be ready for a review. 🎉 

There's quite a bit of changes here, so it might be worth reviewing commit by commit. Here are the highlights.

1. First and foremost I added `Wallaby.Experimental.Selenium`. It has most of the basic test cases working but there are a few things I haven't had time to port over from phantomjs. Those are dialogs, getting the file upload tests to work with selenium and screenshots (I don't think selenium supports them).
2. I set up the build to run the unit tests, and then integration tests for phantom, selenium 2, and selenium 3. It took some tinkering in `start_webdriver.sh` but I think I have a good build environment now. Here's the build matrix on Travis.
screen shot 2017-04-27 at 3 37 02 pm
3. I've gone through and fixed several issues with integration tests that have come up since I started running them against Selenium. The biggest issue was selenium doesn't wait until the page loads (after `visit`, clicking a link, or clicking a button) before it runs the next command. That was giving us a lot of failures when asserting with the `current_url` function. The way I solved this was to create a couple page objects with `ensure_page_loaded/1` functions that use `find/2` to wait until the page has loaded successfully before continuing. I'm sure we can expand the page object paradigm if we want, but this seemed like an ok way to get the tests working.

There's definitely still more to do on this driver (like handling weird responses from BrowserStack), but I am hoping to get this code merged into master so the PR doesn't get huge as I fix more and more issues. I set it to `@moduledoc false` and it's under the `Wallaby.Experimental` namespace, so people should realize that it's not ready for primetime, yet. Please let me know if you'd like any changes and I'd love to get it merged as soon as we can.

keathley:
@aaronrenner Regarding the issue with selenium waiting for page loads we may need to look at this: https://www.w3.org/TR/webdriver/#navigation

integration_test/cases/browser/click_button_test.exs
```diff
@@ -1,6 +1,8 @@
 defmodule Wallaby.Integration.Browser.Actions.ClickButtonTest do
   use Wallaby.Integration.SessionCase, async: true
 
+  alias Wallaby.Integration.PageObjects.IndexPage
```
keathley: Small note but it might be misleading to call these `Objects`. I just call them `Pages` typically.

integration_test/cases/browser/current_path_test.exs
```diff
@@ -1,11 +1,14 @@
...
+  alias Wallaby.Integration.PageObjects.{IndexPage, Page1}
+
   test "gets the current_url of the session", %{session: session}  do
     url =
       session
-      |> visit("")
-      |> click_link("Page 1")
+      |> IndexPage.visit()
+      |> IndexPage.click_page_1_link()
+      |> Page1.ensure_page_loaded()
```
keathley: I love seeing all of these! Using best practices in our own tests is awesome!

integration_test/support/page_objects/index_page.ex
```diff
@@ -0,0 +1,20 @@
...
+  use Wallaby.DSL
+
+  def visit(session) do
+    session
+    |> visit("")
+  end
+
+  def click_page_1_link(session) do
+    session
+    |> click_link("Page 1")
```
keathley: We'll need to update this to use `click(link("Page 1"))`.

lib/wallaby/experimental/selenium.ex
```diff
@@ -0,0 +1,158 @@
...
+  alias Wallaby.Experimental.Selenium.WebdriverClient
+
+  @type start_session_opts ::
+    {:remote_url, String.t} |
+    {:capabilities, map} |
+    {:create_session_fn, ((String.t, map) -> {:ok, %{}})}
+
+  @spec start_session([start_session_opts]) :: {:ok, Session.t}
+  def start_session(opts \\ []) do
+    base_url = Keyword.get(opts, :remote_url, "http://localhost:4444/wd/hub/")
```
keathley: I suspect this option will need to change to something like `:selenium_remote_url` once we add chromedriver support as well.

lib/wallaby/experimental/selenium.ex
```diff
@@ -0,0 +1,158 @@
...
+
+  def execute_script(parent, script, arguments \\ []) do
+    WebdriverClient.execute_script(parent, script, arguments)
+  end
+
+  def send_keys(parent, keys) do
+    WebdriverClient.send_keys(parent, keys)
+  end
+
+  # Unsupported
```
keathley: I think we can remove this.

lib/wallaby/session.ex
```diff
@@ -44,7 +44,7 @@ defmodule Wallaby.Session do
     id: String.t,
     session_url: String.t,
     url: String.t,
-    server: pid(),
+    server: pid | nil,
```
keathley: This is a super small nitpick but could we use an atom here like `:empty` or `:none` or maybe even `:selenium`? Just to signify that we don't actually use the server key for certain drivers.

lib/wallaby/experimental/selenium.ex
aaronrenner: You'd start a session by calling:
```elixir
Wallaby.start_session(driver: Wallaby.Experimental.Selenium, remote_url: "http://localhost:4444/wd/hub")
```

The `:driver` option determines which module to call `start_session/1` on and the rest of the options are passed through directly to the driver. That way chromedriver could look like this:

```elixir
Wallaby.start_session(driver: Wallaby.Experimental.ChromeDriver, remote_url: "http://localhost:4444/wd/hub")
```

integration_test/cases/browser/click_button_test.exs
PragTob: Seconded! :)

aaronrenner:
@keathley I looked into the webdriver `pageLoadStrategy` setting that you mentioned in your link, but I am still getting issues with tests to check the current page using the currentUrl function. I was thinking we could add a function like `IndexPage.loaded?/1`that would find a unique element on the page to ensure the expected page has loaded. Maybe that's something that could be done as part of removing the deprecated `Browser.click_button` work.

Please let me know if there's anything else you'd like to see on this PR. I think it should be good to go as soon as #209 is merged.

lib/wallaby/experimental/selenium.ex
keathley: That makes sense.

aaronrenner:
I think I finally took care of the flake tests and it's ready to merge. I added some retry logic to `SessionCase.start_test_session` and made it so `Selenium.start_session` can better handle error tuples. Let me know if you see anything else I missed.

PragTob:
I was sort of thinking about giving this another look but honestly 1000 lines is a bit more than my casual evening read :D 

If you're both :+1: on it then I'm in as well, otherwise I might get to look at it on the weekend but no promises. :)

aaronrenner:
Sorry it's such a big PR. I'll make my next one much smaller. 😄

keathley:
Ran through it one last time. Looks good to me :+1:

PR #199, opened 2017-04-14
Don't parse JSON on 204 status codes

aaronrenner:
I have seen this issue with BrowserStack returning 204 status codes when calling the /visit command on their Android 4.0 and 4.1 browsers. This change prevents wallaby from crashing because there is no JSON to parse on a 204 status code.

### DIFF
--- a/lib/wallaby/phantom/driver.ex
+++ b/lib/wallaby/phantom/driver.ex
@@ -14,6 +14,11 @@ defmodule Wallaby.Phantom.Driver do
   @type params :: %{using: String.t, value: query}
   @type locator :: Session.t | Element.t
 
+  @type response_errors ::
+    {:error, :invalid_selector} |
+    {:error, :stale_reference_error}
+
+
   @spec create(pid, Keyword.t) :: {:ok, Session.t}
   def create(server, opts) do
     base_url = Wallaby.Phantom.Server.get_base_url(server)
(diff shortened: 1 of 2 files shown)


lib/wallaby/phantom/driver.ex
```diff
@@ -14,6 +14,11 @@ defmodule Wallaby.Phantom.Driver do
   @type params :: %{using: String.t, value: query}
   @type locator :: Session.t | Element.t
 
+  @type response_errors ::
+    {:error, :invalid_selector} |
+    {:error, :stale_reference_error}
+
+
```
PragTob: I guess we wanna get rid off that one extra blank line :)

lib/wallaby/phantom/driver.ex
```diff
@@ -140,11 +145,11 @@ defmodule Wallaby.Phantom.Driver do
   @doc """
   Visits a specific page.
   """
+  @spec visit(Session.t, String.t) :: :ok
   def visit(session, path) do
     check_logs! session, fn ->
-      with {:ok, resp} <- request(:post, "#{session.url}/url", %{url: path}),
-           {:ok, value} <- Map.fetch(resp, "value"),
-        do: {:ok, value}
```
PragTob: somehow astonished `value` wasn't being used apparently :o

lib/wallaby/phantom/driver.ex
aaronrenner: Thanks for pointing that out. Just pushed a fix. 😎

keathley:
Do we wanna return an `:ok` if we get a 204? Or should that be an error?

aaronrenner:
On BrowserStack, 204 means you successfully navigated to a new page, but it just doesn't return a response. I think we should return `:ok` since webdriver did what we asked it to do and we don't care about the response.

keathley:
Makes sense to me :+1:

PR #200, opened 2017-04-18
Extracted integration tests

aaronrenner:
This extracts integration tests (tests that require a webdriver instance) from the unit tests. This allows us to run the same tests against multiple drivers to ensure compatibility. The integration test project structure is based on https://github.com/elixir-ecto/ecto. Below are the important folders/files

* `integration_test/cases` - Integration tests to be run against multiple drivers
* `integration_test/<driver_name>` - Integration tests that will be run for the given driver.
* `integration_test/<driver_name>/all_test.exs` - This file requires the test cases out of `integration_test/cases` that are to be run for this specific browser. Tests have to be manually required out of `integration_test/cases`, but this allows skipping integration tests that aren’t supported by a specific driver.
* `test/` - Unit tests that don’t require a webdriver instance

I did keep the same test folder structure (Browser, Element and Query folders), and we can always rearrange tests at a later date if it makes sense.


coveralls.json
```diff
@@ -0,0 +1,5 @@
+{
+  "skip_files": [
+    "test/support"
+  ]
+}
```
PragTob: :+1:

integration_test/cases/browser/cookies_test.exs
```diff
@@ -15,21 +15,21 @@ defmodule Wallaby.Browser.CookiesTest do
...
-      	|> Browser.set_cookie("api_token", "abc123")
-      	|> visit("/")
-      	|> Browser.cookies
-      	|> hd()
+        session
+        |> visit("/")
+        |> Browser.set_cookie("api_token", "abc123")
+        |> visit("/")
+        |> Browser.cookies
+        |> hd()
```
PragTob: :+1:

integration_test/cases/browser_test.exs
```diff
@@ -0,0 +1,21 @@
+defmodule Wallaby.Integration.BrowserTest do
+  use Wallaby.Integration.SessionCase, async: true
+
+  alias Wallaby.Query
```
PragTob: doesn't `use Wallaby.DSL` already do this?

integration_test/support/session_case.ex
```diff
@@ -0,0 +1,34 @@
+defmodule Wallaby.Integration.SessionCase do
```
PragTob: Not sure if `Session` is still the best name here? I'm also at a loss for a better one, it's more or less a feeling...

integration_test/support/session_case.ex
```diff
@@ -0,0 +1,34 @@
...
+  @doc """
+  Starts a test session with the default opts for the given driver
+  """
+  def start_test_session(opts \\ []) do
+    session_opts = case System.get_env("WALLABY_DRIVER") do
+      "phantom" ->
+        []
+      other ->
+        raise "Unknown value for WALLABY_DRIVER environment variable: #{other}"
+    end |> Keyword.merge(opts)
```
PragTob: Huh, not a fan of the `|> Keyword.merge` after the case... if the case was on its own method it might be more pipable? :)

mix.exs
```diff
@@ -14,13 +16,21 @@ defmodule Wallaby.Mixfile do
      description: "Concurrent feature tests for elixir",
      deps: deps(),
      docs: docs(),
-     test_coverage: [tool: ExCoveralls],
+
+     # Custom testing
+     aliases: ["test.all": ["test", "test.drivers"],
+               "test.drivers": &test_drivers/1],
```
PragTob: Personally I'm a fan that `mix test` executes all tests, as it does now. So, what I'm saying I'd rather have:

`test` --> everything
`test.unit` --> unit test
`test.integration` --> integration tests

dunno how they do it in ecto, but `mix tes t` to test all it can is my general expectation :)

mix.exs
```diff
@@ -65,4 +75,24 @@ defmodule Wallaby.Mixfile do
...
+
+  defp test_paths(driver) when driver in @drivers, do: ["integration_test/#{driver}"]
+  defp test_paths(_), do: ["test"]
+
+  defp test_drivers(args) do
+    for driver <- @drivers, do: run_integration_test(driver, args)
+  end
+
+  defp run_integration_test(driver, args) do
+    args = if IO.ANSI.enabled?, do: ["--color"|args], else: ["--no-color"|args]
```
PragTob: minor nitpick - maybe we should get credo or something, like space around `|` :)

mix.exs
```diff
@@ -65,4 +75,24 @@ defmodule Wallaby.Mixfile do
...
+
+    IO.puts "==> Running tests for WALLABY_DRIVER=#{driver} mix test"
+    {_, res} = System.cmd "mix", ["test"|args],
```
PragTob: space stuff :D

mix.exs
```diff
@@ -65,4 +75,24 @@ defmodule Wallaby.Mixfile do
...
+                          into: IO.binstream(:stdio, :line),
+                          env: [{"WALLABY_DRIVER", driver}]
+
+    if res > 0 do
+      System.at_exit(fn _ -> exit({:shutdown, 1}) end)
+    end
```
PragTob: This all seems really manual :scream: 

I've never done anything like it and I guess you were inspired by ecto and I guess they can be trusted with that, I just wished/thought there would be a more automated/less involved way to do this (not needing to send own System mix commands, but just... invoking another mix task as in "this should also run" a bit more like rake from the ruby world)

integration_test/support/session_case.ex
PragTob: Ok sorry thought there would be now this and another SessionCase, this way it seems good. Sorry :D

mix.exs
aaronrenner: I agree with having `test` run everything and then `test.unit` and `test.integration` run unit and integration tests, respectively. However, I'm worried about overwriting the built-in `test` task. If you have any examples of how to safely improve the names of these tasks, I'd be super happy to make the changes.

mix.exs
aaronrenner: This is the first time I've done this type of integration test setup, and I'll have to admin I copied and pasted from Ecto. 😅 I think the reason it's so manual is because it's adjusting environment variables, which then adjusts the `test_paths:` setting in the project config. We could probably find another way of getting mix to behave the way we want, but I think following Ecto's example isn't a bad for a first try.

integration_test/phantom/all_test.exs
```diff
@@ -0,0 +1,34 @@
...
+Code.require_file "../cases/browser/screenshot_test.exs", __DIR__
+Code.require_file "../cases/browser/select_test.exs", __DIR__
+Code.require_file "../cases/browser/send_text_test.exs", __DIR__
+Code.require_file "../cases/browser/stale_nodes_test.exs", __DIR__
+Code.require_file "../cases/browser/text_test.exs", __DIR__
+Code.require_file "../cases/browser/title_test.exs", __DIR__
+Code.require_file "../cases/browser/visible_test.exs", __DIR__
+Code.require_file "../cases/browser/window_size_test.exs", __DIR__
+Code.require_file "../cases/element/send_keys_test.exs", __DIR__
+Code.require_file "../cases/query_test.exs", __DIR__
```
keathley: It seems like it might be error prone to include these files individually. I can see a scenario where someone forgets to add a new test case to these files. Maybe it makes sense to be selective about which tests we run for each driver?

mix.exs
keathley: I'm OK with it just being `test.all`. Thats similar to the pattern used in `db_connection` as well. Overriding `test` seems like it could have unintended side-effects.

mix.exs
keathley: I agree that this seems manual. I'm OK leaving it for now though.

mix.exs
```diff
@@ -14,13 +16,21 @@ defmodule Wallaby.Mixfile do
...
-     ],
-    dialyzer: [plt_add_apps: [:inets]]
-   ]
+       "coveralls": :test,
+       "coveralls.detail": :test,
+       "coveralls.post": :test,
+       "coveralls.html": :test,
+       "coveralls.travis": :test,
+       "test.all": :test,
+       "test.drivers": :test],
```
keathley: Do you need to include this line. Isn't this already covered by `test.all`?

aaronrenner: It looks like I can remove this line and things will still work when I run `mix test.drivers` (to test all drivers). With this line removed, `mix test.drivers` starts up in the `:dev` environment. Then when `mix test.drivers` spawns each of the driver test suites using `mix test`, those processes will run in the `:test` environment (since they are separate OS processes). I feel like it's kind of unexpected that `mix test.drivers` would run in the `:dev` environment while kicking off tests in that run in the `:test` environment, but it does work. I'm happy to remove this line if you'd like.

integration_test/phantom/all_test.exs
aaronrenner: I can see that scenario too, but this does allow us to skip a test case that's not supported by a specific driver. Maybe another option would be to use a tag based approach (like you do with the pending examples), but I'm trying to figure out what that would look like.

```elixir
test "phantom specific behavior", only_drivers: ["phantom"] do

end
```
or 
```elixir
test "behavior that isn't supported by phantom", except_drivers: ["chrome"] do

end
```

At first glance, I'm not that big a fan of this approach either. Especially if we accidentaly had`only_drivers` and `except_drivers` on the same test case.

Any ideas on how to implement this and still have control on which tests run against which drivers would be appreciated. 🙂

aaronrenner: It looks like db_connection made a [`tests.exs`](https://github.com/elixir-ecto/db_connection/blob/master/integration_test/tests.exs) file that had all of the common requiring code here and then that file is required from [`all_test.exs`](https://github.com/elixir-ecto/db_connection/blob/master/integration_test/sojourn/all_test.exs). Maybe I'll do something like that for now and then we can figure out if we need more flexibility once we have multiple drivers.

aaronrenner: Just pushed an update so we now have a common `tests.exs`.

keathley:
Looks good to me :+1:

PR #201, opened 2017-04-21
Move current driver into Wallaby.Session

aaronrenner:
This moves the responsibility for determining the Wallaby driver to use to the Session and removes all of the hard-coded references to `Wallaby.Phantom.Driver`.

I also created a `Wallaby.Driver` behavior which new drivers will need to implement. I left this behavior with `@moduledoc false` in case we wanted to adjust the interface before making it public. Also, I'm thinking the ideal api for `Wallaby.start_session/1` will be:

```elixir
Wallaby.start_session(driver: Wallaby.Selenium, remote_url: "http://...")
```
instead of 
```elixir
Wallaby.start_session(driver: Wallaby.Selenium.Driver, remote_url: "http://...")
```
which is why I've delegated all of those functions in `Wallaby.Phantom`.

### DIFF
--- a/lib/wallaby/browser.ex
+++ b/lib/wallaby/browser.ex
@@ -988,13 +988,13 @@ defmodule Wallaby.Browser do
     session
   end
 
-  def cookies(%Session{driver: driver}=session) do
+  def cookies(%Session{driver: driver} = session) do
     {:ok, cookies_list} = driver.cookies(session)
 
     cookies_list
   end
 
-  def set_cookie(%Session{driver: driver}=session, key, value) do
+  def set_cookie(%Session{driver: driver} = session, key, value) do
     if blank_page?(session) do
       raise Wallaby.CookieException
     end
@@ -1014,7 +1014,7 @@ defmodule Wallaby.Browser do
   @doc """
   Accepts all subsequent JavaScript dialogs in the given session.
   """
-  def accept_dialogs(%Session{driver: driver}=session) do
+  def accept_dialogs(%Session{driver: driver} = session) do
     driver.accept_dialogs(session)
     session
   end
@@ -1022,7 +1022,7 @@ defmodule Wallaby.Browser do
   @doc """
   Dismisses all subsequent JavaScript dialogs in the given session.
   """
-  def dismiss_dialogs(%Session{driver: driver}=session) do
+  def dismiss_dialogs(%Session{driver: driver} = session) do
     driver.dismiss_dialogs(session)
     session
   end
@@ -1037,7 +1037,7 @@ defmodule Wallaby.Browser do
   end
   ```
   """
-  def accept_alert(%Session{driver: driver}=session, fun) do
+  def accept_alert(%Session{driver: driver} = session, fun) do
     driver.accept_alert(session, fun)
   end
 
@@ -1051,7 +1051,7 @@ defmodule Wallaby.Browser do
   end
   ```
   """
-  def accept_confirm(%Session{driver: driver}=session, fun) do
+  def accept_confirm(%Session{driver: driver} = session, fun) do
     driver.accept_confirm(session, fun)
   end
 
@@ -1066,7 +1066,7 @@ defmodule Wallaby.Browser do
   end
   ```
   """
-  def dismiss_confirm(%Session{driver: driver}=session, fun) do
+  def dismiss_confirm(%Session{driver: driver} = session, fun) do
     driver.dismiss_confirm(session, fun)
   end
 
@@ -1091,15 +1091,15 @@ defmodule Wallaby.Browser do
   end
   ```
   """
-  def accept_prompt(%Session{}=session, fun) do
+  def accept_prompt(%Session{} = session, fun) do
     do_accept_prompt(session, nil, fun)
   end
 
-  def accept_prompt(%Session{}=session, [with: input_value], fun) when is_binary(input_value) do
+  def accept_prompt(%Session{} = session, [with: input_value], fun) when is_binary(input_value) do
     do_accept_prompt(session, input_value, fun)
   end
 
-  defp do_accept_prompt(%Session{driver: driver}=session, input_value, fun) do
+  defp do_accept_prompt(%Session{driver: driver} = session, input_value, fun) do
     driver.accept_prompt(session, input_value, fun)
   end
 
@@ -1113,7 +1113,7 @@ defmodule Wallaby.Browser do
   end
   ```
   """
-  def dismiss_prompt(%Session{driver: driver}=session, fun) do
+  def dismiss_prompt(%Session{driver: driver} = session, fun) do
     driver.dismiss_prompt(session, fun)
   end
 
--- a/lib/wallaby/browser.ex
+++ b/lib/wallaby/browser.ex
@@ -104,7 +104,6 @@ defmodule Wallaby.Browser do
   """
 
   alias Wallaby.Element
-  alias Wallaby.Phantom.Driver
   alias Wallaby.Query
   alias Wallaby.Query.ErrorMessage
   alias Wallaby.Session
@@ -411,10 +410,10 @@ defmodule Wallaby.Browser do
   """
   @spec take_screenshot(parent) :: parent
 
-  def take_screenshot(screenshotable) do
+  def take_screenshot(%{driver: driver} = screenshotable) do
     image_data =
       screenshotable
-      |> Driver.take_screenshot
+      |> driver.take_screenshot
 
     path = path_for_screenshot()
     File.write! path, image_data
@@ -427,8 +426,8 @@ defmodule Wallaby.Browser do
   """
   @spec window_size(parent) :: %{String.t => pos_integer, String.t => pos_integer}
 
-  def window_size(session) do
-    {:ok, size} = Driver.get_window_size(session)
+  def window_size(%Session{driver: driver} = session) do
+    {:ok, size} = driver.get_window_size(session)
     size
   end
 
@@ -437,8 +436,8 @@ defmodule Wallaby.Browser do
   """
   @spec resize_window(parent, pos_integer, pos_integer) :: parent
 
-  def resize_window(session, width, height) do
-    {:ok, _} = Driver.set_window_size(session, width, height)
+  def resize_window(%Session{driver: driver} = session, width, height) do
+    {:ok, _} = driver.set_window_size(session, width, height)
     session
   end
 
@@ -453,8 +452,8 @@ defmodule Wallaby.Browser do
   """
   @spec current_url(parent) :: String.t
 
-  def current_url(parent) do
-    Driver.current_url!(parent)
+  def current_url(%Session{driver: driver} = session) do
+    driver.current_url!(session)
   end
 
   def get_current_url(parent) do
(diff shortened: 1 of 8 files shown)


lib/wallaby/browser.ex
```diff
@@ -1015,16 +1013,16 @@ defmodule Wallaby.Browser do
   @doc """
   Accepts all subsequent JavaScript dialogs in the given session.
   """
-  def accept_dialogs(%Session{}=session) do
-    Driver.accept_dialogs(session)
+  def accept_dialogs(%Session{driver: driver}=session) do
```
PragTob: not for this, but we maybe should get credo going - the narrow spaced assignments bother me  and it's mixed throughout the code base :D

lib/wallaby/element.ex
```diff
@@ -101,27 +100,27 @@ defmodule Wallaby.Element do
...
+        raise Wallaby.StaleReferenceException
     end
   end
 
   @doc """
   Returns a boolean based on whether or not the element is selected.
 
   ## Note
-  This only really makes sense for options, checkboxes, and radi buttons.
+  This only really makes sense for options, checkboxes, and radio buttons.
```
PragTob: :+1: drive by fix

lib/wallaby/phantom/driver.ex
```diff
@@ -37,7 +37,8 @@ defmodule Wallaby.Phantom.Driver do
       session_url: base_url <> "session/#{id}",
       url: base_url <> "session/#{id}",
       id: id,
-      server: server
+      server: server,
+      driver: Wallaby.Phantom
```
PragTob: seems a bit weird that the driver itself doesn't adopt the driver behaviour?

lib/wallaby/phantom.ex
```diff
@@ -50,6 +52,36 @@ defmodule Wallaby.Phantom do
...
+  defdelegate clear(element),                                     to: Driver
+  defdelegate displayed(element),                                 to: Driver
+  defdelegate selected(element),                                  to: Driver
+  defdelegate set_value(element, value),                          to: Driver
+  defdelegate text(element),                                      to: Driver
+
+  defdelegate execute_script(session_or_element, script, args),   to: Driver
+  defdelegate find_elements(session_or_element, compiled_query),  to: Driver
+  defdelegate send_keys(session_or_element, keys),                to: Driver
+  defdelegate take_screenshot(session_or_element),                to: Driver
```
PragTob: all those delegates I'm unsure about... I know it's to support that the interface just says `Wallaby.Phantom` instead of `Wallaby.Phantom.Driver` - personally though I have no problem with it saying `Wallaby.Phantom.Driver` and if I can trade all those delegates against that - then personally I'm more than fine with it :)

lib/wallaby/browser.ex
aaronrenner: I completely agree on credo. I'm just haven't been ready to fix all of the style issues throughout the code-base at the moment.

lib/wallaby/phantom.ex
aaronrenner: Check out the last commit on [this branch](https://github.com/aaronrenner/wallaby/commits/ar-selenium-for-real) to see the what the selenium driver looks like. In that case, I like all of the functions necessary to build a driver being on a single, top-level module. The reason I used delegates was to make `Wallaby.Phantom` conform to the same interface without having to move a ton of code at the moment.

I do realize that I have a lot of functions in my test Selenium driver that could be converted to delegates at the moment
```elixir
def text(%Element{} = element) do
  WebdriverClient.text(element)
end
```
but that's only because I've extracted all of the http calls into a Selenium specific webdriver-client. The interface between `Wallaby.Experimental.Selenium` and `Wallaby.Experimental.Selenium.WebdriverClient` is still fluid and eventually won't match up exactly with what's expected by the `Wallaby.Driver` behaviour. My goal is for `Wallaby.Phantom.Driver` to be replaced by extracting a phantom specific api client, plus moving additional logic like checking logs into `Wallaby.Phantom`.

lib/wallaby/phantom/driver.ex
aaronrenner: `Wallaby.Phantom.Driver` is going to be an internal module that will eventually go away and have it's logic split between to `Wallaby.Phantom` and `Wallaby.Phantom.WebdriverClient`. Check out [my comment below](https://github.com/keathley/wallaby/pull/201#discussion_r112809389) for an example of how I'm doing it with my selenium driver.

lib/wallaby/browser.ex
PragTob: neither should you, was just an aside realization from me :)

lib/wallaby/phantom.ex
PragTob: Hm, I think I see what you're doing. Having all the Driver interaction in a separate module certainly is nice. 

Just seems a bit cumbersome to me to always have to add a delegate to the top level module whenever new functionality is implemented or a new version of a function is added or changed. Plus, when we have the top level module it also seems a bit weird to me that it has a bunch of setup functionality but then delegates all the other stuff. Sort of feels like it is both a high level delegator and a lower level implementor/pool setup. Just thinking out loud here, maybe also another module should handle all the setup so that the top level module is just mostly a high level delegator (save for user agent/capabilities). Just a thought :)

keathley:
Thanks for all of this work @aaronrenner! I think I understand the direction that you're going with this. My only questions are about the distinction between the "driver" and the "webdriverclient". I just want to make sure that we're using the same terms for things :).

Our current phantom setup is like this:

wallaby_supervision_tree

A `Session` holds onto a `Server`. We're now also going to include a `driver` field as part of the session that includes the driver module the session is supposed to be using. These drivers will delegate most of the api calls to a `WebdriverClient`. Based on your PR it seems like you're transitioning from calling the webdriver bits the `Driver` (which is what it is now) to calling the entire supervision tree fragment (top level module, server, and what we now call driver) the "Driver". What we now call a driver would become a `WebdriverClient`. None of this is bad I just want to make sure that we're sharing the same language.

I guess I'm wondering why we wouldn't just set the new `driver` field to the webdriver module. That could be handled during session creation and the end user would never have to know about the name of the webdriver client module. The selenium driver can re-use the `server` field even though there's only going to be one selenium "server". Doing it this way means that we don't have to create the extra delegates and the behaviour could be applied to each driver implementation.

One of the reasons that I like this approach is that it allows us to create multiple driver implementations for each browser. We've talked before about moving away from webdriver in the future as a way to increase performance and stability.maintaining the same api but trying experimental drivers per browser is something I think will be beneficial.

lib/wallaby/browser.ex
aaronrenner: Just checked in a fix to clean up the inconsistent spacing in the params. Thanks for pointing this out. 😄

aaronrenner:
Thanks for the response @keathley. I probably just need to explain my idea for this a little better.

First, a `Driver` would be any module that implements the `Wallaby.Driver` behaviour. Right now, that does happen to map to `Wallaby.Phantom.Driver` and `Wallaby.Selenium.WebDriverClient`, but it doesn't have to (like with non-webdriver based drivers). Any child modules of `Wallaby.Phantom` and `Wallaby.Selenium` are implementation details of that specific driver and the core wallaby code should not interact with it. Wallaby itself should only call functions on the provided `Driver`module.

The reason I pulled out `WebDriverClient` is to make the http interactions easier to test. In `Wallaby.Phantom.Driver`, each call is wrapped with a `check_logs` call that makes additional HTTP calls that are a pain to stub. Same issue happens with the `get_window_size` and `set_window_size` functions that have to first go and first make a separate HTTP request to get the window handle id. I was trying to isolate the http interactions with #195, but I completely understand your point about not being ready to create a common webdriver client that's shared across drivers. Eventually, once we have successful integrations with multiple webdrivers, it might make sense to have some level of common webdriver client, but we can figure it out later. Everything downstream of a Driver module (like `Wallaby.Phantom` or `Wallaby.Selenium`) is an implementation detail and can be refactored and optimized until our heart's content as long as the driver's interface is not broken. That's why for the moment, `Wallaby.Phantom` happens to delegate a lot to `Wallaby.Phantom.Driver`, but in the future maybe `Wallaby.Phantom.Driver` will go away and a function might look like this.

```elixir
def click(%Element{} = element) do
  check_logs fn ->
    WebDriverClient.click(element)
  end
end
```

From an OTP standpoint, servers and supervisor modules will be created the first time the driver's `start_session/1` is called. Here's how it might work in a few examples:

* `Wallaby.start_session(driver: Wallaby.Phantom)` - starts up the pool of phantom servers and returns a session with the `server` property set to one of the pids in the pool. When `Wallaby.end_session/1` is called it will close the session via the webdriver api, and check the server's pid back into the pool.
* `Wallaby.start_session(driver: Wallaby.Phantom, remote_url: "http://example.com")` - Starts a session on a remote phanotmjs server and returns a session with the `server` property set to nil. When `Wallaby.end_session/1` is called, it will close the session via the webdriver api (no need to check in a server).
* `Wallaby.start_session(driver: Wallaby.Selenium)` - starts up a single selenium server and starts a session via the webdriver api. When `Wallaby.end_session/1` is called, it will close the session via the webdriver api. There's probably no reason to have the `server` property set, because I don't see any logic to shut down the wallaby server when all of the sessions are closed.
* `Wallaby.start_session(driver: Wallaby.Selenium, remote_url: "http://browserstack.com/wd/hub")` - Starts a session on a remote selenium server and returns a session with the `server` property set to nil. When `Wallaby.end_session/1` is called it will close the session via the webdriver api (no need to check in a server).

If you happen to be using both `Wallaby.Phantom` and `Wallaby.Selenium` at the same time, each driver will start it's own OTP hierarchy (if necessary) when `Wallaby.start_session` is called with its respective driver name.

Thanks again for your feedback. Please let me know what you think about this design, or if you'd like me to explain certain aspects more. I definitely want to help prepare wallaby to accept as many drivers as possible. Also, regarding naming, would it make sense to rename `Driver` to `Adapter` since driver is so close to webdriver?

keathley:
Thanks @aaronrenner! That helps clarify a lot of things. Overall I think this seems like a reasonable way to move forward. I thought about changing `Driver` to something like `Adapter` or `Browser`. In the end I opted for calling the user api the `Browser` and the underlying driver implementations "drivers". I'm bad at naming things :)

I'll go through this one more time and see if I pick up on anything but I think we should be good to move forward unless you have any thoughts @PragTob.

lib/wallaby/driver.ex
```diff
@@ -0,0 +1,158 @@
...
+
+  @doc """
+  Invoked to retrieve cookies for the given session.
+  """
+  @callback cookies(Session.t) :: {:ok, [%{String.t => String.t}]} | {:error, reason}
+
+  @doc """
+  Invoked to get the current path of the browser's session.
+  """
+  @callback current_path!(Session.t) :: String.t | nil | no_return
```
keathley: This doesn't need to be fixed in this PR but I think we should remove these in favor of always using `{:ok, result} | {:error, reson}` in the driver modules.

PragTob:
Sounds good to me too, thank you @aaronrenner for taking the time to implement and also explain your ideas so eloquently :+1: 

I like the idea of Wallaby only talking to the top level module. That gives a nice separation of concerns and we could actually extract the drivers to their own projects if maturity level X is reached. Drivers and improvements could then also be shared with hound or others that wanna use them independently :)

PragTob:
I guess with the ok from @keathley we are good to go now and I'll get the honor to be the one to press the magic button :D 

Thanks @aaronrenner ! :heart:

PR #203, opened 2017-04-26
Allow adjusting hackney options

aaronrenner:
I was having an issue where BrowserStack wasn't sending me a response on some of my commands, causing my script to just hang. This change will allow me to override the hackney options and set custom timeouts.

### DIFF
--- a/README.md
+++ b/README.md
@@ -301,6 +301,25 @@ Application.put_env(:wallaby, :js_logger, file)
 
 Logging can be disabled by setting `:js_logger` to `nil`.
 
+## Config
+
+### Adjusting timeouts
+
+Wallaby uses [hackney](https://github.com/benoitc/hackney) under the hood, so we
+offer a hook that allows you to control any hackney options you'd like to have
+sent along on every request. This can be controlled with the `:hackney_options`
+setting in `config.exs`.
+
+```elixir
+# default values
+config :wallaby,
+  hackney_options: [timeout: :infinity, recv_timeout: :infinity]
+
+# Overriding a value
+config :wallaby,
+  hackney_options: [timeout: 5_000]
+```
+
 ## Contributing
 
 Wallaby is a community project. PRs and Issues are greatly welcome.
--- a/config/config.exs
+++ b/config/config.exs
@@ -7,7 +7,8 @@ config :wallaby,
   pool_size: 3,
   js_logger: :stdio,
   screenshot_on_failure: false,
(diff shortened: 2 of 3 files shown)


aaronrenner:
Actually, now that I think about it, maybe this setting should be on the driver level.

```elixir
config :wallaby, Wallaby.Phantom, 
  hackney_options: [timeout: :infinity, recv_timeout: :infinity]
```

What do you think?

keathley:
Setting it per driver seems like a good call to me.

README.md
```diff
@@ -301,6 +301,25 @@ Application.put_env(:wallaby, :js_logger, file)
 
 Logging can be disabled by setting `:js_logger` to `nil`.
 
+## Config
+
+### Adjusting timeouts
+
+Wallaby uses [hackney](https://github.com/benoitc/hackney) under the hood, so we
```
PragTob: technically it uses httpoison which uses hackney, should be fine though - was just a bit confusing to me.

README.md
```diff
@@ -301,6 +301,25 @@ Application.put_env(:wallaby, :js_logger, file)
...
+setting in `config.exs`.
+
+```elixir
+# default values
+config :wallaby,
+  hackney_options: [timeout: :infinity, recv_timeout: :infinity]
+
+# Overriding a value
+config :wallaby,
+  hackney_options: [timeout: 5_000]
```
PragTob: I think this overrides the complete options (not just a value)? Or does config do a merge of some kind on the values? I think it doesn't do merging...

PragTob:
Haha and there we disagree :D I'd still like to get the reasoning, but am fine with it as you two seem to be in agreement :)

keathley:
@PragTob I could go either way on it. The only reason I thought breaking it out per driver might be useful is so each driver could specify their own timeouts. I'm not sure that makes sense or not at this point. I'm totally fine with a single config for now.

README.md
aaronrenner: I tried out this behavior locally using `IO.pry/1` and confirmed my project config setting was merged in with the defaults provided in Wallaby's `config.exs`.

aaronrenner:
@PragTob I think you're right that we don't need per-driver configuration. Unless you'd like any other changes, this should be good to merge, as-is.

PragTob:
:shipit:

aaronrenner:
Thanks @PragTob. 👍

PR #204, opened 2017-04-26
Only document modules for public API

aaronrenner:
This removes quite a few modules from the API docs that aren't meant for public usage. This will allow us to know exactly what functions people depend on and which ones we need to consider backward compatibility on. Here's what the API docs look like now.

screen shot 2017-04-26 at 1 59 30 pm

There's still quite a few of deprecated functions in `Wallaby.Browser` that need to be removed, but I figured they will be taken care of as part of #177.

### DIFF
--- a/lib/wallaby.ex
+++ b/lib/wallaby.ex
@@ -19,6 +19,7 @@ defmodule Wallaby do
 
   alias Wallaby.Session
 
+  @doc false
   def start(_type, _args) do
     import Supervisor.Spec, warn: false
 
@@ -36,7 +37,35 @@ defmodule Wallaby do
     {atom, any}
 
   @doc """
-  Starts a browser session
+  Starts a browser session.
+
+  ## Multiple sessions
+
+  Each session runs in its own browser so that each test runs in isolation.
+  Because of this isolation multiple sessions can be created for a test:
+
+  ```
+  @message_field Query.text_field("Share Message")
+  @share_button Query.button("Share")
+  @message_list Query.css(".messages")
+
+  test "That multiple sessions work" do
+    {:ok, user1} = Wallaby.start_session
+    user1
+    |> visit("/page.html")
+    |> fill_in(@message_field, with: "Hello there!")
+    |> click(@share_button)
+
+    {:ok, user2} = Wallaby.start_session
+    user2
+    |> visit("/page.html")
+    |> fill_in(@message_field, with: "Hello yourself")
+    |> click(@share_button)
+
+    assert user1 |> find(@message_list) |> List.last |> text == "Hello yourself"
+    assert user2 |> find(@message_list) |> List.first |> text == "Hello there"
+  end
+  ```
   """
   @spec start_session([start_session_opts]) :: {:ok, Session.t} | {:error, reason}
   def start_session(opts \\ []) do
@@ -45,25 +74,29 @@ defmodule Wallaby do
   end
 
   @doc """
-  Ends a browser session
+  Ends a browser session.
   """
   @spec end_session(Session.t) :: {:ok, Session.t} | {:error, reason}
   def end_session(%Session{driver: driver} = session) do
     driver.end_session(session)
   end
 
+  @doc false
   def screenshot_on_failure? do
     Application.get_env(:wallaby, :screenshot_on_failure)
   end
 
+  @doc false
   def js_errors? do
     Application.get_env(:wallaby, :js_errors, true)
   end
 
+  @doc false
   def js_logger do
     Application.get_env(:wallaby, :js_logger, :stdio)
   end
 
+  @doc false
   def phantomjs_path do
     Application.get_env(:wallaby, :phantomjs, "phantomjs")
   end
--- a/lib/wallaby/dsl.ex
+++ b/lib/wallaby/dsl.ex
@@ -1,5 +1,30 @@
 defmodule Wallaby.DSL do
-  @moduledoc false
+  @moduledoc """
+  Sets up the Wallaby DSL in a module.
+
+  All functions in `Wallaby.Browser` are now accessible without a module name
+  and `Wallaby.Browser`, `Wallaby.Element` and `Wallaby.Query` are all aliased.
+
+  ## Example
+
+  ```
+  defmodule MyPage do
+    use Wallaby.DSL
+
+    @name_field Query.text_field("Name")
+    @email_field Query.text_field("email")
+    @save_button Query.button("Save")
+
+    def register(session) do
+      session
+      |> visit("/registration.html")
+      |> fill_in(@name_field, with: "Chris")
+      |> fill_in(@email_field, with: "c@keathly.io")
+      |> click(@save_button)
+    end
+  end
+  ```
+  """
 
   defmacro __using__([]) do
     quote do
--- a/lib/wallaby/helpers/key_codes.ex
+++ b/lib/wallaby/helpers/key_codes.ex
@@ -1,8 +1,8 @@
 defmodule Wallaby.Helpers.KeyCodes do
-  @moduledoc """
-  Helper utility for converting key atoms into key codes sutiable to send over
-  the wire.
-  """
+  @moduledoc false
+
+  # Helper utility for converting key atoms into key codes suitable to send over
+  # the wire.
 
   @doc """
   Encode a list of key codes to a usable json representation.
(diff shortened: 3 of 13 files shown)


aaronrenner:
I also ran across [this link](https://hexdocs.pm/elixir/1.4.2/writing-documentation.html#hiding-internal-modules-and-functions) in the Elixir docs about hiding modules that aren't part of the public api.

lib/wallaby.ex
```diff
@@ -36,7 +37,35 @@ defmodule Wallaby do
...
+
+    {:ok, user2} = Wallaby.start_session
+    user2
+    |> visit("/page.html")
+    |> fill_in(@message_field, with: "Hello yourself")
+    |> click(@share_button)
+
+    assert user1 |> find(@message_list) |> List.last |> text == "Hello yourself"
+    assert user2 |> find(@message_list) |> List.first |> text == "Hello there"
+  end
```
PragTob: Code examples :heart_eyes:

lib/wallaby/dsl.ex
```diff
@@ -1,5 +1,26 @@
...
+  @moduledoc """
+  Sets up the Wallaby DSL in a module.
+
+  All functions in `Wallaby.Browser` are now accessible without a module name
+  and `Wallaby.Browser`, `Wallaby.Element` and `Wallaby.Query` are all aliased.
+
+  ## Example
+
+  ```
+  defmodule MyPageObject do
```
PragTob: I know that coming from Ruby one tends to call them page objects, but they aren't really as it's just a module. I would just call it `MyPage` - what do you think?

lib/wallaby/dsl.ex
```diff
@@ -1,5 +1,26 @@
...
+
+    def register(session) do
+      session
+      |> visit("/registration.html")
+      |> fill_in(Query.text_field("Name"), with: "Chris")
+      |> fill_in(Query.text_field("email"), with: "c@keathly.io")
+      |> click(Query.button("Save"))
+    end
+  end
+  ```
```
PragTob: Thanks for adding these comprehensive examples! :+1:

lib/wallaby/helpers/key_codes.ex
```diff
@@ -1,8 +1,5 @@
 defmodule Wallaby.Helpers.KeyCodes do
-  @moduledoc """
-  Helper utility for converting key atoms into key codes sutiable to send over
-  the wire.
-  """
+  @moduledoc false
```
PragTob: While we already have the comment, maybe we keep it as a code comment? I'm not a big fan of those, but a high level comment for the module might help for new contributors? Wdyt?

lib/wallaby/metadata.ex
```diff
@@ -1,9 +1,5 @@
 defmodule Wallaby.Metadata do
-  @moduledoc """
-  Metadata is used to encode information about the browser and test. This
-  information is then stored in a User Agent string. The information from the
-  test can then be extracted in the application.
-  """
```
PragTob: While we already have the comment, maybe we keep it as a code comment? I'm not a big fan of those, but a high level comment for the module might help for new contributors? Wdyt?

lib/wallaby/phantom.ex
```diff
@@ -1,13 +1,45 @@
...
+
+  Start a Wallaby Session using this driver with the following command:
+
+  ```
+  {:ok, session} = Wallaby.start_session(driver: Wallaby.Phantom)
+  ```
+
+  ## Notes
+
+  This driver requires PhantomJS be installed in your path. You can install PhantomJS through NPM or your package manager of choice:
```
PragTob: unless you specify the phantomjs path (which is what I usually do, all to the same npm package so everyone runs the same version) - maybe we could make that clearer here?

lib/wallaby/phantom.ex
```diff
@@ -41,63 +75,96 @@ defmodule Wallaby.Phantom do
...
+  @doc false
   defdelegate text(element),                                      to: Driver
 
+  @doc false
   defdelegate execute_script(session_or_element, script, args),   to: Driver
+  @doc false
   defdelegate find_elements(session_or_element, compiled_query),  to: Driver
+  @doc false
   defdelegate send_keys(session_or_element, keys),                to: Driver
+  @doc false
```
PragTob: a `doc @false start/end` would be useful here :D

lib/wallaby/dsl.ex
keathley: I second that idea.

lib/wallaby/dsl.ex
```diff
@@ -1,5 +1,26 @@
...
+  defmodule MyPageObject do
+    use Wallaby.DSL
+
+    def register(session) do
+      session
+      |> visit("/registration.html")
+      |> fill_in(Query.text_field("Name"), with: "Chris")
+      |> fill_in(Query.text_field("email"), with: "c@keathly.io")
+      |> click(Query.button("Save"))
```
keathley: One other addition that would be good to add here is to move these queries into module attributes or functions. Thats a pattern that I think we should be actively trying to encourage. What do y'all think?

lib/wallaby/metadata.ex
keathley: Whats the standard here? I'm OK with leaving these sorts of docs simply because they help other contributors.

lib/wallaby/phantom.ex
keathley: We could update the example below this to show how to use the phantom in your node modules

lib/wallaby/session.ex
```diff
@@ -1,44 +1,5 @@
 defmodule Wallaby.Session do
-  @moduledoc """
```
keathley: Glad this is finally being removed 👍

lib/wallaby/metadata.ex
PragTob: According to the link aaron posted to the elixir docs using code comments for private functions is the standard:

> ##  Documenting Private Functions
>
> Elixir warns if a private function has a @doc attribute and discards its content, because @doc is intended to be used only for your public interface.
>
> Private functions may still need internal documentation for maintainers, though. That can be accomplished with code comme

lib/wallaby/dsl.ex
PragTob: Totally :+1: from my side

lib/wallaby/dsl.ex
aaronrenner: That sounds good. I'll check in a commit shortly.

lib/wallaby/metadata.ex
aaronrenner: I felt bad deleting that documentation, so I think a code comment is a great idea. 👍

lib/wallaby/phantom.ex
aaronrenner: @PragTob I'm not sure I understand what these docs should look like. I copied these notes from the readme and have an example of setting the `:phantomjs` config value on line 24. What would be a good way to improve the clarity here?

aaronrenner:
I just checked in the suggested updates, but wasn't sure what to do for [doc changes](https://github.com/keathley/wallaby/pull/204#discussion_r113864849) in `lib/wallaby/phantom.ex`.

lib/wallaby/dsl.ex
aaronrenner: I just updated this to use module attributes and match the rest of the documentation. I'm not sure why, but dialyzer throws a fit when I store these queries in module attributes in my personal project. It could be just me, though.

```
def set_email(session, email) do
  session
  |> fill_in(Query.css("[name=email]"), with: email)
end
```
works, but 
```
@email_field Query.css("[name=email]")

def set_email(session, email) do
  session
  |> fill_in(@email_field, with: email)
end
```
gives me the following warning
```
Function set_email/2 has no local return
```

lib/wallaby/phantom.ex
PragTob: I'd say that it requires it in the path or you configure a custom path. As for the example I think it's something like `node_modules/bin/phantomjs` or something :)

PragTob:
Just commented on those :)

aaronrenner:
Just updated the example phantomjs path to point to `node_modules/.bin/phantomjs`. I think I've incorporated all of the feedback so far. Let me know if you'd like any more changes.

PragTob:
Well let's get this party in then! Thanks! :)

PR #209, opened 2017-05-01
Fix leaking webdriver sessions in QueryTest

aaronrenner:
`Wallaby.Integration.QueryTest` was starting a ton of sessions and not cleaning them up. This switches it over to use `Wallaby.Integration.SessionCase` and uses the session that's automatically created and cleaned up. I also extracted the unit tests out of this integration test so they can be run separately.

I'm also thinking this should help the reliability of my tests for #198.

### DIFF
--- a/integration_test/cases/query_test.exs
+++ b/integration_test/cases/query_test.exs
@@ -1,15 +1,8 @@
 defmodule Wallaby.Integration.QueryTest do
-  use ExUnit.Case, async: true
+  use Wallaby.Integration.SessionCase, async: true
   use Quixir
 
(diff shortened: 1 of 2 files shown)


integration_test/cases/query_test.exs
```diff
@@ -1,15 +1,8 @@
 defmodule Wallaby.Integration.QueryTest do
-  use ExUnit.Case, async: true
+  use Wallaby.Integration.SessionCase, async: true
   use Quixir
```
keathley: I had so many good intentions of property testing the query generation process

PR #210, opened 2017-05-02
Automatically clean up sessions created with start_test_session/1

aaronrenner:
There's a few changes in here:

1. Added an integration test that ensures the driver allows calling `Wallaby.end_session/1` on a closed session. There was an issue in the phantom driver where the check_logs call would fail during end_session, so I added an option to skip checking logs. (I was getting a 404 `Variable Resource Not Found` response from phantom when it tried to check logs while clearing local storage during shutdown. I'm wondering if there's some async stuff going on here and if shutdown got in the way).
2. Set up an `on_exit` callback directly in `start_test_session`, so sessions created through this function are automatically shut down when the test exits.
3. Added a tag to `Wallaby.Integration.SessionCase` that allows us to be able to skip automatically generating a test session.


keathley:
We can remove the check_log call in end session. That would probably fix the errors with checking logs on a closed session. #95 will take care of this long term.

PR #214, opened 2017-05-09
Remove deprecated version of find.

keathley:
This PR removes the deprecated version of find.


keathley:
This is related to #177.

PragTob:
I do wonder if those were really all invocations because truth be told I was dreading `find` a bit :D Looks great to me, thanks @keathley ! :)

PR #215, opened 2017-05-09
Remove deprecated version of text

keathley:
Removes the deprecated version of text. If you need to get the text of an Element you can do so by specifying `Element.text`.

I did question whether the session version of the function should be on `Browser` or `Element`. I see arguments for either. I left it on `Browser` for now but I'm open to other ideas.


keathley:
Related to #177

PragTob:
Leaving it on browser seems good to me so :+1:

PR #219, opened 2017-05-10
Remove deprecated versions of all

keathley:
Just like the title says. Removes the deprecated version of all.


keathley:
Related to #177

PR #225, opened 2017-05-11
Fix test case for a button with no type via text

britto:
In the [fixtures](https://github.com/keathley/wallaby/blob/81d69ea/integration_test/support/pages/forms.html#L43-L44), this button's text is actually _button with no type_. The current text (_Submit button_) selects an element which belogs to [another](https://github.com/keathley/wallaby/blob/81d69ea/integration_test/cases/browser/click_button_test.exs#L48) test case: _clicking button type[submit] via button text_.


keathley:
Good catch @britto :+1:

PR #226, opened 2017-05-12
moving logging to server

### LINKED ISSUE #95: Capture JS errors and logs in the servers `handle_info` call

krankin:
this attempts to resolve #95.


krankin:
Woops.. this isn't ready.

PR #228, opened 2017-05-12
Fix bad button type error reporting.

keathley:
If there are duplicate buttons then Wallaby will incorrectly return an error saying that the button has a bad type. But the real error is that the specified count is incorrect. This PR adds a test for this case and fixes the underlying bug.


PragTob:
And another conflict... I don't believe it :|

PragTob:
And finally this can go in :)

PR #229, opened 2017-05-12
Remove deprecated version of visible?

keathley:
This PR removes the deprecated version of `visible?`.

### DIFF
--- a/integration_test/cases/browser/visible_test.exs
+++ b/integration_test/cases/browser/visible_test.exs
@@ -7,26 +7,26 @@ defmodule Wallaby.Integration.Browser.VisibleTest do
     test "determines if the element is visible to the user", %{page: page} do
       page
       |> find(Query.css("#visible"))
-      |> visible?
+      |> Element.visible?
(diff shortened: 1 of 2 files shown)


integration_test/cases/browser/visible_test.exs
```diff
@@ -7,26 +7,26 @@ defmodule Wallaby.Integration.Browser.VisibleTest do
...
+      assert Element.visible?(element) == false
     end
 
     @tag skip: "Unsuported in phantom"
     test "handles obscured elements", %{page: page} do
       element = find(page, Query.css("#obscured", visible: false))
 
-      assert visible?(element) == false
+      assert Element.visible?(element) == false
     end
```
aaronrenner: I just tried this obscured test in selenium and it doesn't work either. Is it worth removing?

PragTob: Hm I guess so then, I don't like dead code :)

Maybe better in a separate PR though?

PR #235, opened 2017-05-12
Report duplicate html issues correctly.

keathley:
This is similar to the issue with finding duplicate buttons. We're incorrectly reporting malformed labels when its really an issue with finding an incorrect amount of elements. This is noticeable when finding radio buttons and checkboxes.


PragTob:
:tada:

PR #239, opened 2017-05-13
Changelog updates

PragTob:
I thought I'd clean up the the Changelog a bit/add the bunch of PRs I just merged right in there so it's more up to date and ready to go.

Other comments/ideas/things I did that we might have differing opinions on inline :)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## 0.17.0 (unreleased)
 
+This release removes all methods declared as _deprecated_ in the 0.16 release, experimental Selenium support and much more! If you are looking to upgrade from an earlier release, it is recommended to first go to 0.16.x.
+Other goodies include improved test helpers, a cookies API and handling for JS-dialogues.
+
+
 ### Breaking Changes
 
 * Removed deprecated version of `fill_in`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,18 +34,14 @@ Other goodies include improved test helpers, a cookies API and handling for JS-d
 * Ability to pass options for driver interaction down to the underlying hackney library through `config :wallaby, hackney_options: [your: "option"]` - thanks @aaronrenner
 * Added `check_log` option to `execute_script` - thanks @aaronrenner
 * Experimental support for selnium 2 and selenium 3 web drivers has been added, use at your own risk ;)
+* Updated hackney and httpoison dependencies - thanks @aaronrenner
+* Removed documentation for modules that aren't intended for external use - thanks @aaronrenner
 
 ### Bugfixes
 
 * Fix spawning of phantomjs when project path contains spaces - thanks @schnittchen
 * Fixed a couple of dialyzer warnings - thanks @aaronrenner
-* Fixed incorrect malformed label warning when it was rreally a mismatch between expected elements found
-
-### Chores
-
-* Updated hackney and httpoison dependencies - thanks @aaronrenner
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -1,16 +1,39 @@
 ## 0.17.0 (unreleased)
 
+This release removes all methods declared as _deprecated_ in the 0.16 release, experimental Selenium support and much more! If you are looking to upgrade from an earlier release, it is recommended to first go to 0.16.x.
+Other goodies include improved test helpers, a cookies API and handling for JS-dialogues.
+
+
+### Breaking Changes
```
PragTob: I added `Breaking Changes` because imo these are neither Features nor Bugfixes and often also the most important things when I look at a Changelog

CHANGELOG.md
```diff
@@ -1,16 +1,39 @@
...
 * New cookie API with `cookies/1` and `set_cookie/3`
 * New assert macros `assert_has/2` and `refute_has/2`
 * execute_script now returns the session again and is pipable, there is an optional callback if you need access to the return value - thanks @krankin
-* Removed deprecated version of fill_in
 * Phantom server is now compatible with escripts - thanks @aaronrenner
 * Ability to handle JavaScript dialogs via `accept_dialogs/1`, `dismiss_dialogs/1`, plus methods for alerts, confirms and prompts - thanks @padde
 * Ability to pass options for driver interaction down to the underlying hackney library through `config :wallaby, hackney_options: [your: "option"]` - thanks @aaronrenner
 * Added `check_log` option to `execute_script` - thanks @aaronrenner
-* Removed `check`
+* Experimental support for selnium 2 and selenium 3 web drivers has been added, use at your own risk ;)
```
PragTob: I think that this is such a big and great thing, although it is experimental it should totally be mentioned in the Changelog. Wdyt? @aaronrenner

PragTob:
Argh I can't comment so far down.

I'd wanna remove `## Chores` and move them either to Feature/Bugfix/Breaking Changes and if not applicable remove them. I know chores often take up a lot of work so it's sad not to have them in the Changelog but I think the Changelog should only have information that is important to the end users of a library. E.g. no internal refactorings or things that were fixed within a release (like we introduce a bug, but fix it before a version with that bug ever goes live).

Wdyt @keathley @aaronrenner ? :)

keathley:
Removing `## Chores` is totally fine with me :+1:

CHANGELOG.md
aaronrenner: I think the error handling needs work some work, making it not a friendly experience for someone new to the library. It's not documented yet, and I'm worried about the number of issues that would be opened if we announced it before it's documented and is friendlier to the user.

PragTob: Hm.. I sort of still would wanna mention it, how many people read the Changelog and then try it out even if we say at own risk? Probably more than I like, sure. I'm a bit more worried about the amount of people that goes "Oh wallaby still doesn't support selenium... are they even working on it?"

If you still feel uncomfortable with it, sure we should take it out - mostly your changes :)

keathley: Personally I'm OK with mentioning the selenium driver. Its definitely marked as experimental so folks would have to go out of their way to use it. If they use it and open issues then thats kinda awesome because it means its something that people want, they took the time to actually try it out and they care enough to open an issue about it. The issues can help us prioritize the remaining work and track down bugs. Really the worst case scenario is that no one uses it :)

PR #240, opened 2017-05-15
Remove select and selected?/1 from Browser module.

keathley:
This PR removes the `select` functions and `selected?/1` from the Browser module. It would be very nice to have some sort of query composition for finding options inside of a given select. Because there are multiple "Option 2" options on the page you get a duplicate element error if you don't scope everything to the specific select.

### DIFF
--- a/integration_test/cases/browser/select_test.exs
+++ b/integration_test/cases/browser/select_test.exs
@@ -1,7 +1,7 @@
 defmodule Wallaby.Integration.Browser.SelectTest do
   use Wallaby.Integration.SessionCase, async: true
 
-  import Wallaby.Query, only: [css: 1]
+  import Wallaby.Query, only: [option: 1, select: 1]
 
   setup %{session: session} do
     page =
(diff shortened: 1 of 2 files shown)


integration_test/cases/browser/select_test.exs
```diff
@@ -11,84 +11,26 @@ defmodule Wallaby.Integration.Browser.SelectTest do
...
 
   describe "selected?/2" do
     test "returns a boolean if the option is selected", %{page: page} do
       assert page
-      |> find(Query.select("My Select"))
-      |> select(Query.option("Option 2"))
-      |> selected?(Query.option("Option 2")) == true
+      |> find(select("My Select"))
+      |> click(option("Option 2"))
+      |> selected?(option("Option 2")) == true
```
aaronrenner: Why does this test call `selected/2` and the one below calls `Element.selected/2`? Also would it be good to have a test to ensure this function returns `false` when the option is not selected?

keathley: The test below should be calling `Element.selected?/1`. I'll add a test to make sure that it returns false. Thats a good catch.

keathley:
@aaronrenner Added the extra tests. Lemme know what you think.

aaronrenner:
Thanks, the tests look good. Is the reason wallaby has `Browser.selected` and `Element.selected` is because `Browser.selected` has retry logic and `Element.selected` does not?

keathley:
Partially. In general the `Browser` module provides functions that accept queries and provide retrying, etc. The `Element` module accepts elements and doesn't provide any retry logic. In fact the `Element` module really _can't_ provide any retry logic. Typically if a query needs to be retried its because an element wasn't able to be found or it *was* found and then went stale half-way through the element validation (checking for text or visibilty).

aaronrenner:
That makes sense. Thanks for the explanation.

PR #245, opened 2017-05-17
Update CHANGELOG.md for 0.17 release

keathley:
I'm not actually sure what the protocol for updating the changelog should be. If there is a better way to manage this then please let me know :)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 0.17.0
+## 0.17.0 (2017-05-17)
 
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -1,4 +1,4 @@
-## 0.17.0 (unreleased)
+## 0.17.0
```
PragTob: I think we should put a release date. Like 2017-05-17 :)

keathley:
Awesome! Sounds good to me.

PR #246, opened 2017-05-17
Add :get option to Queries to select one element from a list

graeme-defty:
After much consideration I selected `get` as the keyword to use. I am not completely happy with this choice, but all the other options I considered sounded worse.

I made the numbering 1-based. I have no problem selecting the 0th element of an array while programming, but in discussing links on a screen I think it is more intuitive for users to refer to the second link by `get: 2`.

### DIFF
--- a/config/docs.exs
+++ b/config/docs.exs
@@ -1,0 +1,1 @@
+use Mix.Config


PragTob:
:wave: 

I thin there are a couple of canges that don't belong here due to git management.

It's often preferable to don't do Pull Requests from master but keep master of a fork in sync with the original repository (e.g. this one). Then you can create branches to implement the features with the branch really just having that one change.

But how do you get master back in sync?

I have 2 remotes setup, my own fork and wallaby itself. 

```
tobi@speedy ~/github/wallaby $ git remote -v
origin	git@github.com:keathley/wallaby.git (fetch)
origin	git@github.com:keathley/wallaby.git (push)
own	git@github.com:PragTob/wallaby.git (fetch)
own	git@github.com:PragTob/wallaby.git (push)
```

You could try something like:

```
git branch -D master # deletes your master branch
git fetch origin # should be the remote that is this wallaby repo
git checkout -b master origin/master # should check out the main wallaby I hope
git push -f # hopefully overrides the master of your own repo
git checkout -b my-awesome-get-branch # create the new branch
```

I think/hope this'll work, there might be a better way but this seems solid to me :)

lib/wallaby/browser.ex
```diff
@@ -857,6 +856,16 @@ defmodule Wallaby.Browser do
     end
   end
 
+  defp do_get(query, elements) do
+    getNumber = query.conditions[:get]
+    case {getNumber, length(elements)} do
+        {nil,_}     -> {:ok, elements}
+        {n, count} when n > 0 and n <= count
```
PragTob: this looks like 4 space indention, we prefer 2 spaces :)

lib/wallaby/browser.ex
```diff
@@ -857,6 +856,16 @@ defmodule Wallaby.Browser do
     end
   end
 
+  defp do_get(query, elements) do
+    getNumber = query.conditions[:get]
```
PragTob: please use snake_case :)

lib/wallaby/query.ex
```diff
@@ -235,13 +236,17 @@ defmodule Wallaby.Query do
     Keyword.get(conditions, :count)
   end
 
+  def getNumber(%Query{conditions: conditions}) do
```
PragTob: please `get_number` although not sure that'ts the best name but I can't come up with a better atm

PragTob: Also as it's a public method I'd like to see a little unit test :)

lib/wallaby/browser.ex
PragTob: Also I think it'd be better if we use the `get_number` method from the Query module here, to not bind this to the implementation details of where this comes from :)

graeme-defty:
Thanks for the advice (see my previous statements about git!)

I did do the development on a separate branch this time, but did the merge
into master before I pushed it to Github.

I will try your suggestion and see if it does the trick.

If not, I am more than happy to delete my fork and re-fork it.

Thanks again (for advice AND patience)

g

PragTob:
@graeme-defty NP :) Merging master in before is a good idea, but just on your branch. E.g. on the branch do `git merge master` or even better `git rebase master` :)

graeme-defty:
Yes, I did that (or tried to!)

So then I push with the changes still in my branch only, yes?

PragTob:
@graeme-defty yes just your branch :) Wgen you do this here we'll have to create a new PR but that shouldn't be a problem :)

integration_test/cases/query_test.exs
```diff
@@ -70,6 +70,31 @@ defmodule Wallaby.Integration.QueryTest do
...
+
+  test "queries can not select an element off the start of the list", %{session: session} do
+    assert_raise Wallaby.QueryError, fn ->
+      session
+      |> Browser.visit("/page_1.html")
+      |> Browser.find(Query.css(".user", count: 5, get: 0))
+    end
+  end
+
+  test "queries can not select an element off the end of the list", %{session: session} do
```
graeme-defty: changed this to "past"

lib/wallaby/query.ex
graeme-defty: Name changed.

However, I could find no tests for similar functions (count, inner_text, et al) so I followed suit.

graeme-defty:
I think all is fixed now.

Thanks to Ghost for his sage advice!

PR #249, opened 2017-05-22
Add 'at' to query

graeme-defty:
Hopefully incorporating all requested changes and also nicely gitted


.gitignore
```diff
@@ -3,6 +3,7 @@
 /deps
 erl_crash.dump
 *.ez
+mix.lock
```
PragTob: Still not wanting the mix.lock in gitignore :)

integration_test/cases/query_test.exs
```diff
@@ -70,6 +70,31 @@ defmodule Wallaby.Integration.QueryTest do
...
+    end
+  end
+
+  test "queries can not select an element off the end of the list", %{session: session} do
+    assert_raise Wallaby.QueryError, fn ->
+      session
+      |> Browser.visit("/page_1.html")
+      |> Browser.find(Query.css(".user", count: 5, get: 6))
+    end
+  end
```
PragTob: :+1: good boundary checking testing!

lib/wallaby/browser.ex
```diff
@@ -857,6 +857,16 @@ defmodule Wallaby.Browser do
     end
   end
 
+  defp do_get(query, elements) do
+    getNumber = query.conditions[:get]
+    case {getNumber, length(elements)} do
```
PragTob: for variable names we prefer `snake_case` over `camelCase` (maybe that's not camel but something differnet, you know what I'm gettign at) :)

lib/wallaby/browser.ex
```diff
@@ -857,6 +857,16 @@ defmodule Wallaby.Browser do
...
+        {nil,_}     -> {:ok, elements}
+        {n, count} when n > 0 and n <= count
+                    -> {:ok, [Enum.at(elements, n-1)]}
+        {_, _}      -> {:error, {:get_number, length(elements)}}
```
PragTob: smaller style pet peeve(not too important and our fault for not having credo setup) is that I like all case statements to be one line or all of them to be multiline - not a biggie though :)

lib/wallaby/query.ex
```diff
@@ -235,13 +236,17 @@ defmodule Wallaby.Query do
     Keyword.get(conditions, :count)
   end
 
+  def getNumber(%Query{conditions: conditions}) do
```
PragTob: again `snake_case` vs. `otherCase` :)

lib/wallaby/browser.ex
```diff
@@ -857,6 +857,16 @@ defmodule Wallaby.Browser do
     end
   end
 
+  defp do_get(query, elements) do
+    getNumber = query.conditions[:get]
```
PragTob: as you have the `getNumber` (soon `get_number`) in the query module, I'd prefer to use that so that we can encapsulate that detail there :)

lib/wallaby/query.ex
```diff
@@ -29,6 +29,7 @@ defmodule Wallaby.Query do
     * `:count` - The number of elements that should be found (default: 1).
     * `:visible` - Determines if the query should return only visible elements (default: true).
     * `:text` - Text that should be found inside the element (default: nil).
+    * `:get` - The number of the element to return if multimple elements satisfy the selection criteria.
```
PragTob: not sure if `get` is the best name, but atm I fail to come up with something better... @keathley ? :)

lib/wallaby/query.ex
```diff
@@ -289,4 +295,8 @@ defmodule Wallaby.Query do
         |> Keyword.put_new(:maximum, opts[:maximum])
     end
   end
+
+  defp add_get(opts) do
+    Keyword.put_new(opts, :get, nil)
```
PragTob: Do we always have to add it? Ah I guess that's the pattern the code above follows so - good for following the established patterns! :+1:

lib/wallaby/query.ex
```diff
@@ -235,13 +236,17 @@ defmodule Wallaby.Query do
...
+  end
+
   def inner_text(%Query{conditions: conditions}) do
     Keyword.get(conditions, :text)
   end
 
   def result(query) do
     cond do
-      count(query) == 1 ->
+      count(query) == 1 || getNumber(query) != nil ->
```
PragTob: I know it's a small condition, but I find myself looking at this and not really understanding it until I put some solid thought into it. So, personally I'd like to extract the second condition into something like `specific_element_requested` or something? :)

mix.lock
```diff
@@ -18,15 +18,15 @@
...
+  "plug": {:hex, :plug, "1.3.4", "b4ef3a383f991bfa594552ded44934f2a9853407899d47ecc0481777fb1906f6", [:mix], [{:cowboy, "~> 1.0.1 or ~> 1.1", [hex: :cowboy, repo: "hexpm", optional: true]}, {:mime, "~> 1.0", [hex: :mime, repo: "hexpm", optional: false]}], "hexpm"},
+  "poison": {:hex, :poison, "3.1.0", "d9eb636610e096f86f25d9a46f35a9facac35609a7591b3be3326e99a0484665", [:mix], [], "hexpm"},
+  "pollution": {:hex, :pollution, "0.9.0", "2d172aeedc444f0fc06a69b03bb6a560fb141bcafcc0c8189c29b10ee6e50893", [:mix], [], "hexpm"},
+  "poolboy": {:hex, :poolboy, "1.5.1", "6b46163901cfd0a1b43d692657ed9d7e599853b3b21b95ae5ae0a777cf9b6ca8", [:rebar], [], "hexpm"},
+  "quixir": {:hex, :quixir, "0.9.1", "b9a45930f330ba485c1cb976afc9f5ceb14ebbe10faf755e0796bb971396c37c", [:mix], [{:pollution, "~> 0.9", [hex: :pollution, repo: "hexpm", optional: false]}], "hexpm"},
+  "ranch": {:hex, :ranch, "1.3.2", "e4965a144dc9fbe70e5c077c65e73c57165416a901bd02ea899cfd95aa890986", [:rebar3], [], "hexpm"},
+  "ssl_verify_fun": {:hex, :ssl_verify_fun, "1.1.1", "28a4d65b7f59893bc2c7de786dec1e1555bd742d336043fe644ae956c3497fbe", [:make, :rebar], [], "hexpm"},
```
PragTob: not sure where all those changes come from, but I'm gonna guess `hex` client update so sounds fine ti me to commit them if they don't cause merge conflicts  otherwise we should do them in a separate PR :)

.gitignore
graeme-defty: Oh cripes! - I am so sorry. I really thought I had addressed this and all your other comments. I think GITting clumsiness has tripped me up again. :-(

integration_test/cases/query_test.exs
graeme-defty: Agreed. Done.

lib/wallaby/browser.ex
graeme-defty: Agreed. Done.

lib/wallaby/browser.ex
graeme-defty: Agreed. Done.

lib/wallaby/browser.ex
graeme-defty: Personally I think this is clearer, but project standard rules, so ... Done.

lib/wallaby/query.ex
graeme-defty: Same. I thought of pick, choose, return, select and a few others, but in the end this was no worse in meaning than the others, and also the shortest.

lib/wallaby/query.ex
graeme-defty: Agreed. Done.

lib/wallaby/query.ex
graeme-defty: Agreed. Done.
Actually I moved the whole condition into the function because asking for an element which has only one instance is just another way of asking for a specific element, and this makes the `cond` easier to read.
Not sure of the best location for the function but in the end put it immediately after its use.

mix.lock
graeme-defty: Is this a result of my messing about with `mix.lock` in `gitignore`, perhaps?

lib/wallaby/query.ex
PragTob: Hm, I guess I like `choose` and `pick` better, `get` seems a wee bit to generic for me, might also be because getting a specific radio button is called `choose` in capybara but that might also get confusing...

mix.lock
PragTob: No, more likely just a new hex version, it added stuff in the end. new hex versions sometimes do that :)

graeme-defty: Got it ;-)

lib/wallaby/query.ex
graeme-defty: The problem with all these options is they could conceivably refer to a quantity rather than a specific instance.  `index` sounds a bit too 'techie' IMHO. (I still cling to the quaint notion that non-programmers could code tests.  How about `sequence`? Could be shortened to `seq` to help the lazy typists.

PragTob: Not a big fan of lazy typists over here, code is written once but read many times so i favor the expressiveness :) Let's wait a bit of time to see if keath gets time to check in here and add his voice :)

For me `get` reminds me a bit too much of HTTP GET requests which confuses me in thsi context.

graeme-defty: Sorry - I should perhaps not have been flippant about laziness - but brevity IS a virtue. This is not code that will be written once, but a parameter name that will be written (hopefully) thousands of times. All else equal, pick the short one!  A shorter name just may help avoid tests spilling onto another line, and so improve readability.

OK - last offer - how about `item`? It reads well (item 5), carries the air of singularity that most of these suggestions lack AND ... is short. ;-)

Looking forward to keathley's view.

keathley: Out of all of these options I think `item` or `pick` are the ones that make sense to me. I've also thought that it could just be called `nth` since thats how it works in css. Not sure if thats  the most descriptive but it would at least be familiar to folks familiar with css and it describes what you're going to do fairly well.

lib/wallaby/query.ex
keathley: Yeah, all of the options get defaulted when a query is "built".

lib/wallaby/query.ex
graeme-defty: Yes, nth occurred to me since that is what I am used to using in Erlang (though not Elixir). However, again, it seemed a bit 'technical'. My issue with `pick`, along with all my other sugegstions, is that it does not nicely capture the fact that we are selecting only 1 item from the list. (If I was to present someone with a row of chocolates (beers, whatever) and said `pick 3` I doubt I would get the result we are talking about here!)

If you are OK with `item` then I am convinced that is the best.

keathley: Also slight typo here: `multimple` -> `multiple`

lib/wallaby/query.ex
```diff
@@ -289,4 +299,8 @@ defmodule Wallaby.Query do
         |> Keyword.put_new(:maximum, opts[:maximum])
     end
   end
+
+  defp add_get(opts) do
+    Keyword.put_new(opts, :get, nil)
```
keathley: I'd like to avoid using `nil` here. What do you think about defaulting this to something like `:all`. That way its more descriptive. We can pattern match on that in `result/1`

lib/wallaby/query/error_message.ex
```diff
@@ -47,6 +47,11 @@ defmodule Wallaby.Query.ErrorMessage do
     can't apply both filters correctly.
     """
   end
+  def message(_, {:get_number, size}) do
```
keathley: We should pass the query in instead of size here. The query will have the result on it so we can get the size. The query will also have the requested item so we can change this message to something like:

```
The element at index #{query.get} is not available because #{result_count(query.result)} #{method(query)} #{result_expectation(query.result}
```

lib/wallaby/browser.ex
```diff
@@ -857,6 +857,17 @@ defmodule Wallaby.Browser do
     end
   end
 
+  defp do_get(query, elements) do
+    case {Query.get_number(query), length(elements)} do
+      {nil,_} ->
+        {:ok, elements}
+      {n, count} when n > 0 and n <= count ->
```
keathley: I'm not sure how I feel about starting at 1 here. The programmer in me wants to 0-index everything.

keathley:
One other thought I just had was that we could name this option `at`. That would be consistent with Enum, its brief, and it explains exactly what it does.

PragTob:
:+1: for `:at` !

lib/wallaby/query.ex
PragTob: also again here, I'm :+1: for `at`!

Thanks for all the discussions!

graeme-defty: Done

graeme-defty: Done

lib/wallaby/query/error_message.ex
graeme-defty: Yes, good idea. Done

aaronrenner:
Just triggered a rebuild and now travis is passing again.

PragTob:
Other PR merged, this still looks good but waiting for Keath a bit. Also, didn't we wanna switch to `:at` instead of `:item` or did I miss some important discussion?

lib/wallaby/browser.ex
graeme-defty: I did put a bit of thought into this, so let me try to explain my reasoning...

(But before I begin, perhaps we can let your inner programmer sleep easy at nights knowing that many programming languages use 1-based indexing. I have programmed in a few of them myself - including, interestingly enough, Erlang, which is mainly what I have been using for the last 6 years of so.

Maybe that is what is influencing me, but I think it goes rather deeper than that. Another example is css, which was the inspiration for your `nth` suggestion. In css, `nth` is 1-based, and I think this reflects a much wider usage. The use of 0-based is in truth not an _index_, but an _offset_. The beauty of offsets in a programming language is that you can add them and you will get the expected result. But what we are dealing with here is not a programming language but a DSL, and the domain under consideration is a web page. Picking the 3rd item in a list by specifying that you want item at position 2 I would think is counter-intuitive to most people. Instances of 0-based addressing outside of programming languages are very rare indeed. I can not think of a single one, off the top of my head. Certainly there is nobody living at number 0 on my street!

There may be all kinds of technical people who are not programmers - web designers, perhaps, or maybe full-time testers - who end up using Wallaby. I certainly hope so, and they are the ones I am thinking about. It is to make this technology accessible to such people is one of the main reasons that we have DSLs in the first place.

Programmers will deal with it whichever we choose - especially if they have experience of both systems - but for 'normal' people 1-based is the natural way.

So have I done a good enough selling job? I hope so. But, as always, this is your gig, and the final call is yours.

lib/wallaby/query.ex
graeme-defty: Indeed!  Done

graeme-defty: Looking at this is making me like `item` again. `at: :all` makes some kind of odd logic, but `item: :all` just reads so much better.

graeme-defty:
Yes indeed. It is in - progress. You just caught me mid-change.

keathley:
@graeme-defty I totally understand and respect where you're coming from on the indexing. Let me think about it for a bit. Right now my thinking is that if we're using `:at` to be consistent with Enum then we should be consistent with Enum's indexing as well. If we were using `:nth` then it would make more sense to start at 1 since that would be consistent with css. That said, I prefer `:at`  and I'm not advocating that we change the keyword.

As much as I would love to think that lots of other people will use Wallaby, currently we're still trying to make a foothold in the elixir community. Once we have that I'll feel more comfortable making decisions to try to cross the chasm to another niche.

Like I said I'll think about it. But right now I'm leaning towards staying consistent with `Enum.at`. Thanks for putting a lot of thought into this :smile:. I really appreciate it :heart:

keathley:
Sorry for delaying this so long @graeme-defty. Life gets in the way some times. This feature is something that would still be great to have. We've talked about it and I think that for now we'd like to stay consistent with `Enum.at` and just use 0 indexing for everything. If you wanna make those changes then that would be awesome but I totally understand if you don't have the time/desire to finish this up at the moment. If thats the case then I'm happy to finish up the work and get this merged. Thanks for all the effort and sorry again for the long delay :heart:.

keathley:
I've rebased this branch so it'll merge cleanly here: #323.

PR #257, opened 2017-05-31
Resolve issue with filling form fields based on labels where 1 label is subset of another

### LINKED ISSUE #256: Wallaby.Query.text_field - label text is an 'include' query. Fails with 2 fields with same string
My registration form has Password, and Password confirmation.

`fill_in(text_field("Password confirmation"), with: "password")`

works just fine.  However:

`fill_in(text_field("Password"), with: "password")`

fails because it finds 2 elements.  This happens because the generated XPath is doing a text include query.  I can certainly see why such a decision was made, but I think it might be better to bracket that regex in ^\s+ and \s+$.

Will submit a PR.

jeffdeville:
Matching the text of an inner element forces the inner text to match exactly, and be surrounded by nothing but optional whitespace

**NOTE** I had to remove a browser test, so this is a breaking change.  However, I suspect the use case I am solving for is the more common one.  The test in question was searching for text that was a subset of the text in the returned element.

I think that if the previous use case is important, it might be a good idea to add a callback to the list of Query conditions that would allow the user to handle the filtering on their own. 

fixes #256

### DIFF
--- a/integration_test/cases/browser/find_test.exs
+++ b/integration_test/cases/browser/find_test.exs
@@ -86,11 +86,6 @@ defmodule Wallaby.Integration.Browser.FindTest do
       assert user1 != user2
     end
 
-    test "can be scoped by inner text when there are multiple elements with text", %{page: page} do
-      element = find(page, Query.css(".inner-text", text: "Inner Text"))
-      assert element
-    end
-
     test "scoping with text escapes the text", %{page: page} do
       assert find(page, Query.css(".plus-one", text: "+ 1"))
     end
--- a/lib/wallaby/browser.ex
+++ b/lib/wallaby/browser.ex
@@ -870,7 +870,7 @@ defmodule Wallaby.Browser do
   defp matching_text?(%Element{driver: driver} = element, text) do
     case driver.text(element) do
       {:ok, element_text} ->
-        element_text =~ ~r/#{Regex.escape(text)}/
+        element_text =~ ~r/^\s*#{Regex.escape(text)}\s*$/
       {:error, _} ->
         false
     end


keathley:
Hey @jeffdeville thanks for looking into this!

Unfortunately this fix doesn't actually solve your original bug. We don't use `matching_text?` for finding text_fields with text. All of that logic is done with xpath queries in the `Query.Xpath` module. I was able to re-create your original issue by running this test on your branch:

```html
    <div class="login-form">
      <label for="password">Password</label>
      <input type="text" id="password" name="user[password]">

      <label for="password-confirmation">Password Confirmation</label>
      <input type="text" id="password-confirmation" name="user[password-confirmation]">
    </div>
```

```elixir
  test "finds inputs if there are similar labels", %{page: page} do
    assert page
    |> find(css(".login-form"))
    |> fill_in(text_field("Password"), with: "password")
  end
```

In order to fix this we'll need to update those xpath queries to not use a more explicit matching then `contains`.

keathley:
This error:

```
** (Wallaby.QueryError) The text 'Psswrd' matched a label but the label's 'for' attribute
     doesn't match the id of any element.

     Make sure that id on your element is `id="psswrd"`.

     stacktrace:
       (wallaby) lib/wallaby/browser.ex:466: Wallaby.Browser.find/2
       (wallaby) lib/wallaby/browser.ex:445: Wallaby.Browser.find/3
       integration_test/cases/browser/fill_in_test.exs:83: (test)
```

Is being misreported due to the regex change. Under the hood wallaby is finding multiple elements when it only expects to find one. It then tries to see if the html is malformed. Because the semantics between the xpath query and the regex are different now its reporting that the html is broken instead of the real issue which is that there are duplicate elements. In order to fix this will have to update the xpath query for fillable_fields.

In general I'm somewhat against making this change to the regex. It breaks the ability to specify text with css selectors which is a common pattern. Often the css selector is going to contain much more text then what would be specified in a query. We'll need to find a way to solve the root cause without breaking that functionality.

jeffdeville:
Fair enough.  There's probably an argument to be made that the capybara notion of finding a form field with the text label is too magical (and brittle) anyway. It might be 'better' (and certainly faster) to use CSS selectors.  Any further desired abstraction could be provided using a page object (ala site_prism), which provides a useful abstraction for non-trivial applications anyway.

As for changing your XPath, I'm certainly no XPath guy, but it seems like your only conceivable alternative is startsWith, and that'd just push the problem somewhere else. 

Thanks for taking a look. Let me know if I can be of any assistance.

keathley:
You can match exact text in the xpath query. Since its just looking for label text then it shouldn't cause issues since folks typically fully qualify those queries. If you want to take a look at the xpath query module it'll involve changing the `contains(#{query})` calls to something like `text()=#{query}`.

keathley:
If you want to take a stab at it I'm happy to help you out. If not thats fine. We'll prioritize the issue for an upcoming release.

keathley:
I've opened an explicit issue for this bug. I'm going to close this PR for now. @jeffdeville if you'd like to take a look at this fix then I'd be happy to chat with you about it. I'm @keathley on irc and slack if you wanna reach out :+1:.

jeffdeville:
Thanks @keathley. I'm afraid I was bounced to another project and am deep in React ~~hell~~ land at the moment. Thanks for driving this issue forward for me. I much prefer elixir, so hopefully I'll be able to come back to this soon.

keathley:
No worries. Good luck in React land. I've been there a time or two myself and I know the feeling ;). 

Actually I'm very interested to see if there's a way to utilize wallaby outside of standard elixir projects. If you have time to chat about your use case still feel free to reach out to me :+1:.

Issue #174, opened 2017-03-29
Sessions not being cleaned up correctly when tests end


keathley:
Unless the session is explicitly "ended" when the test ends the sessions in phantom are being left open. I discovered this when my phoenix server started receiving duplicate messages from 2 browsers. What was happening is that the first test ran, connected to the channel, and executed the test. When the test ended the phantom was being checked back in but the session wasn't being deleted from the phantom. You normally wouldn't see this sort of issue. It only cropped up because we drive several user actions from external devices and those devices broadcast their data over a separate phoenix channel to all connected clients (which in practice is only ever 1 client).

You can fix this issue by explicitly calling `Wallaby.end_session(session)` in your setup block like this:

```elixir
setup %{session: session} do
  on_exit fn ->
    Wallaby.end_session(session)
  end
end
```

But, we should account for this in Wallaby so the end user doesn't have to care about these things. My initial thinking is that we can set up a monitor or a link to the calling process when it checks out a wallaby session. Then when we detect a `DOWN` signal from the process we can clean up the session ourselves. We'll also need to handle the case where we get a down signal but the user has manually checked in the process.

brandonparsons:
I'm getting some strange behavior /w Wallaby and PhantomJS, and wondering if it is related to this issue.

After running my tests (even if I've excluded all feature tests), I see the following output in the console:

```
Finished in 0.1 seconds
1 test, 0 failures

Randomized with seed 100114
/bin/sh: line 10:  4732 Killed: 9               "$@"    <---------- This only shows up after Wallaby installed
```

And then after the test suite has exited, I end up with a pile of phantomJS processes still running:

```
$ ps aux ....

brandon           5047   0.0  0.3  3674612  43052   ??  S    11:16AM   0:19.34 /Users/brandon/.nvm/versions/node/v6.9.1/lib/node_modules/phantomjs-prebuilt/lib/phantom/bin/phantomjs --webdriver=49414 --local-storage-path=/var/folders/4h/4f2vml1952g4brhb9f9y65wr0000gn/T/tva1ml
brandon           5046   0.0  0.3  3674128  43636   ??  S    11:16AM   0:19.34 /Users/brandon/.nvm/versions/node/v6.9.1/lib/node_modules/phantomjs-prebuilt/lib/phantom/bin/phantomjs --webdriver=49412 --local-storage-path=/var/folders/4h/4f2vml1952g4brhb9f9y65wr0000gn/T/1l1hl82
[...]

PR #279, opened 2017-07-18
Monitor and demonitor sessions for all browsers.

### LINKED ISSUE #269: Wallaby does not close Chrome after tests - using Selenium
I'm using Chrome with Selenium and Wallaby does not close the browsers after the tests finish running.

I have `selenium-server` running and I start my test suit with `WALLABY_DRIVER=selenium mix test`.

This is how I start sessions: 

```elixir
{:ok, session} = Wallaby.start_session(
  capabilities: %{browserName: "chrome"}
)
```

I've tried to call `Wallaby.end_session(opts)` on `on_exit` but ended up facing DB ownership issues.

keathley:
This is the same logic that we use to kill chrome sessions. I've generalized it and made it work for all browsers. The solution is to monitor the processes the starts a session. If we detect a DOWN signal from that process then we kill the session ourselves. If the process calls `end_session` then we demonitor the process and allow wallaby to close the session.

This should fix #269 and #174.

### DIFF
--- a/integration_test/support/session_case.ex
+++ b/integration_test/support/session_case.ex
@@ -20,8 +20,7 @@ defmodule Wallaby.Integration.SessionCase do
       |> Keyword.merge(opts)
 
     with {:ok, session} <- retry(2, fn -> Wallaby.start_session(session_opts) end),
-        :ok <- on_exit(fn -> Wallaby.end_session(session) end),
-        do: {:ok, session}
+      do: {:ok, session}
   end
 
   @doc """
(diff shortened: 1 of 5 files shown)


keathley:
Its also occurred to me that the extra retry logic we added for chromedriver might have also fixed the issues we were seeing when trying to run concurrent tests against a single chromedriver. I removed the pooling logic to see if the issue was resolved and it seems like it is! I was seeing some broken tests in CI and I believe that they were due to pooling issues. Plus removing the need for pooling should free up the system and remove some extra complexity.

PR #284, opened 2017-07-20
Add the correct error message for text queries

### LINKED ISSUE #283: No function clause matching in Wallaby.Query.ErrorMessage.method/2
When I call `assert_has(session, Query.text("my text"))` and Wallaby doesn't find an element with the text, I see this error:

```
** (FunctionClauseError) no function clause matching in Wallaby.Query.ErrorMessage.method/2
     stacktrace:
       (wallaby) lib/wallaby/query/error_message.ex:81: Wallaby.Query.ErrorMessage.method(:text, false)
       (wallaby) lib/wallaby/query/error_message.ex:65: Wallaby.Query.ErrorMessage.found_error_message/1
       (wallaby) lib/wallaby/query/error_message.ex:12: Wallaby.Query.ErrorMessage.message/2
```

When I change the call to:

```elixir
# the same xpath used by Query.text
assert_has(xpath(".//*[contains(normalize-space(text()), 'my text')]"))
```

I see the expected error message:

```
** (Wallaby.ExpectationNotMet) Expected to find 1, visible element that matched the xpath './/*[contains(normalize-space(text()), 'my text')]' but 0, visible elements were found.
```

Seems like `lib/wallaby/query/error_message.ex` is lacking these: 

```elixir
def method(:text, true), do: "texts"
def method(:text, false), do: "text"
```

keathley:
We didn't have a matching function for displaying text query errors.

This fixes #283

### DIFF
--- a/lib/wallaby/query/error_message.ex
+++ b/lib/wallaby/query/error_message.ex
@@ -113,6 +113,8 @@ defmodule Wallaby.Query.ErrorMessage do
   def method(:file_field, true), do: "file fields"
   def method(:file_field, false), do: "file field"
 
+  def method(:text, true), do: "elements with the text"
+  def method(:text, false), do: "element with the text"
 
   def short_method(:css, count) when count > 1,  do: "elements"
   def short_method(:css, count) when count == 0, do: "elements"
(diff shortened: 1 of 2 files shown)


PragTob:
hm looks we couldn't correcly get chromedriver in the one build:

```
The command "curl -L -O "http://chromedriver.storage.googleapis.com/${CHROMEDRIVER_VERSION}/chromedriver_linux64.zip"" failed and exited with 18 during .
```

I'll try rerunning them

Issue #277, opened 2017-07-17
Handle alerts in chrome


keathley:
We need to support the same alert functionality that we use for phantomjs in chrome.

tk8817:
@florinpatrascu - feel like picking this up for fun? we handle this gracefully in `watir` http://watir.github.io/docs/javascript-dialogs/

keathley:
We had to build an api for making this work in phantom. It should be much easier with chrome.

florinpatrascu:
trying to get my head wrapped around this :)

keathley:
@florinpatrascu awesome thanks! If you have specific questions feel free to ask them here or on the elixir slack/irc. I'm `keathley` on both.

florinpatrascu:
I am quite new to this, learning to deal with the chromedriver (in Elixir/Wallaby). My team here is using it in heavy-duty mode for many projects (Ruby/Watir), but we also have newer projects where we're starting to use Wallaby :)

keathley:
Oh rad! Just let me know if I can help in any way.

keathley:
Closing this now that #285 has been merged.

PR #285, opened 2017-07-21
Handle alerts in chrome #277

florinpatrascu:
Hi Chris,

As per our chat earlier, I Implemented the following:

```elixir
accept_alert/2, dismiss_alert/2, accept_confirm/2,
dismiss_confirm/2, accept_prompt/3, dismiss_prompt/2
```

And the following tests, are all passing:

```sh
WALLABY_DRIVER=chrome mix test integration_test/cases/browser/dialog_test.exs:96
WALLABY_DRIVER=chrome mix test integration_test/cases/browser/dialog_test.exs:111
WALLABY_DRIVER=chrome mix test integration_test/cases/browser/dialog_test.exs:139
WALLABY_DRIVER=chrome mix test integration_test/cases/browser/dialog_test.exs:164
``` 
However, there is a caveat: 

 - can't open a dialog/alert/prompt in ChromeDriver 2.30.477690, when in headless mode (OSX here). Probably it works,  if using Xvfb ... but it sucks :/
 
The tests above and the implementation so far, work with the :chrome option [headless: false]

Can't test the whole `dialog_test.exs` suite until we clarify the future functionality of `accept_dialogs/1` and  `dismiss_dialogs/1`, as today these two are very phantomjs-oriented :) Have the code for chrome, but it is incompatible with the current tests, as we discussed.

I believe chrome (2.30) doesn't love me anymore.

### DIFF
--- a/lib/wallaby/experimental/selenium/webdriver_client.ex
+++ b/lib/wallaby/experimental/selenium/webdriver_client.ex
@@ -365,9 +365,7 @@ defmodule Wallaby.Experimental.Selenium.WebdriverClient do
     }
   end
 
-  @doc """
-  Retrieves the text from an alert, prompt or confirm.
-  """
+  # Retrieves the text from an alert, prompt or confirm.
   @spec alert_text(Session.t) :: {:ok, String.t}
   defp alert_text(session) do
     with  {:ok, resp} <- request(:get, "#{session.url}/alert_text"),
--- a/lib/wallaby/experimental/selenium/webdriver_client.ex
+++ b/lib/wallaby/experimental/selenium/webdriver_client.ex
@@ -59,6 +59,54 @@ defmodule Wallaby.Experimental.Selenium.WebdriverClient do
   end
 
   @doc """
+  Invoked to accept one alert triggered within `open_dialog_fn` and return the alert message.
+  """
+  def accept_alert(session, fun) do
+    fun.(session)
+    with  {:ok, value} <- alert_text(session),
+          {:ok, _r} <- request(:post, "#{session.url}/accept_alert"),
+      do: value
+  end
+
+  @doc """
+  Invoked to accept one alert triggered within `open_dialog_fn` and return the alert message.
+  """
+  def dismiss_alert(session, fun) do
+    accept_alert(session, fun)
+  end
+
+  def accept_confirm(session, fun) do
+    accept_alert(session, fun)
+  end
+
+  def dismiss_confirm(session, fun) do
+    fun.(session)
+    with  {:ok, value} <- alert_text(session),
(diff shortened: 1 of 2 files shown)


keathley:
This is looking good to me.

@PragTob, @aaronrenner I've discussed this with @florinpatrascu but I'm wondering what direction we should go with handling alerts across all browsers. Right now the alert functionality is built to accommodate phantom due to phantom's inability to handle js alerts. Functions like `accept_dialogs` and `dismiss_dialogs` are designed to insert some javascript into the browser so that all dialogs can be captured or skipped. I think what I would like to do is to move `accept_dialogs` and `dismiss_dialogs` into the `Browser` module and implement them there. The other dialog and alert functions can continue to live in the driver modules. 

Another improvement I would like to make (unrelated to this PR though) is to have the driver modules return `{:ok, any()} | {:error, any()}` instead of just returning the message.

What do y'all think about this?

lib/wallaby/experimental/selenium/webdriver_client.ex
```diff
@@ -316,4 +364,15 @@ defmodule Wallaby.Experimental.Selenium.WebdriverClient do
       driver: Wallaby.Experimental.Selenium,
     }
   end
+
+  @doc """
+  Retrieves the text from an alert, prompt or confirm.
+  """
+  @spec alert_text(Session.t) :: {:ok, String.t}
+  defp alert_text(session) do
```
keathley: We either need to make this public or remove the doc. Otherwise we get a compiler warning which is currently whats breaking your build. I'm ok with making it public if we want to do that.

florinpatrascu: I'll comment the doc and leave it there in case the method is becoming public, somehow in the future?! 

Re:
> Another improvement I would like to make (unrelated to this PR though) is to have the driver modules return {:ok, any()} | {:error, any()} instead of just returning the message.

I too prefer the `{:ok|:error, .. `, but in this particular round, I was trying to stay as close as possible to the phantom driver implementation, as I am still new to this code - I know you previous comment is not strictly related to this specific PR, but I'd thought I'd state my opinion too, in support of yours :)

florinpatrascu:
I was hoping the new chrome driver (2.31) will let us open dialogs in headless mode; alerts, for example. I probably have to hope more ...

Trying to open an alert box, in headless mode, will return this: `(Session info: headless chrome=59.0.3071.115)\n  (Driver info: chromedriver=2.31.488774 (7e15618d1bf16df8bf0ecf2914ed1964a387ba0b),platform=Mac OS X 10.11.6 x86_64)`

:/

florinpatrascu:
👍

PR #286, opened 2017-07-23
Travis Updates

PragTob:
* OTP 20!
* New Elixir 1.4!
* elixir 1.5 rc!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,16 @@
 language: elixir
 elixir:
-  - 1.4.2
+  - 1.4.5
   - 1.3.4
+  - 1.5.0-rc.2
 otp_release:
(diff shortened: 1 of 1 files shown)


PragTob:
well.. this didn't go terribly well. Mixture of errors from not being able to get chromedriver or whatever, some errors in 1.5.. fun stuff.

keathley:
Looks like we're getting compiler errors with elixir 1.5. Not sure whats up with that.

PragTob:
I'll remove 1.5 for now from here and make a new PR with 1.5 so we can investigate there or locally what's up with that, and maybe put an emergency stop to the release. Damn I should do it on my other projects... oh but blog post etc. :D

PR #287, opened 2017-07-25
Elixir 1.5.0 rc2 try

PragTob:
(will most likely fail, see if we can fix it otherwise tell the good folks at elixir)

### DIFF
--- a/.tool-versions
+++ b/.tool-versions
@@ -1,2 +1,2 @@
-elixir 1.4.5
+elixir 1.5.0
 erlang 20.0
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ language: elixir
 elixir:
   - 1.4.5
   - 1.3.4
-  - 1.5.0-rc.2
+  - 1.5.0
 otp_release:
   - 19.3
   - 18.3
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ language: elixir
 elixir:
   - 1.4.5
   - 1.3.4
+  - 1.5.0-rc.2
 otp_release:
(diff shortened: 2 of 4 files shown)


PragTob:
Woopsie:

```
.** (ExUnit.DuplicateDescribeError) describe "find/3" is already defined in Wallaby.Integration.Browser.FindTest

    integration_test/cases/browser/find_test.exs:130: (module)

    (stdlib) erl_eval.erl:670: :erl_eval.do_apply/6

    (elixir) lib/code.ex:376: Code.require_file/2

    integration_test/chrome/all_test.exs:15: (file)

    (elixir) lib/code.ex:376: Code.require_file/2

    (elixir) lib/kernel/parallel_require.ex:59: anonymous fn/2 in Kernel.ParallelRequire.spawn_requires/5

07:34:18.887 [error] GenServer #PID<0.263.0> terminating

** (ExUnit.DuplicateDescribeError) describe "find/3" is already defined in Wallaby.Integration.Browser.FindTest

    integration_test/cases/browser/find_test.exs:130: (module)

    (elixir) src/elixir_compiler.erl:85: :elixir_compiler.dispatch/6
```

keathley:
That...is a very interesting error. Not sure what changed in the compiler to start throwing those warnings.

PragTob:
I guess they just made duplicated `describe` fail like it is with `test` which makes sense but can't find it in the changelog.

keathley:
Interesting about the duplicate `describe` block. I guess that ends up being helpful if you use the descriptions to find tests. I typically just use line numbers which is probably why I missed it 😈

PragTob:
Ah finally green... on to the merge!

PR #288, opened 2017-07-26
Reduce number of travis builds

PragTob:
Right now we generate [40](https://travis-ci.org/keathley/wallaby/builds/257458811) travis builds. That is a bit too much, I don't want my Travis friends to come over and be like "Tobi, you are using up all our resources!"
On a more serious note, we should fairly use our given free resources. Also, we sometimes have weird troubles downloading drivers etc. (which we should try to fix) but the more builds we have the more likely this is to occur and then we gotta restart etc... also, of course - it takes quite long.

So, of all the things we test what can we let go off?

* We need to test all drivers
* we need to test all elixir versions
* We need to test all combinations of drivers and elixir versions as they might execute different code from our driver integration and I wanna avoid issues where we start calling methods that don't exist yet in 1.3 for instance 

So, what can we do? Well my suggestion is this: Run each elixir version with its "dedicated" OTP release:
* 1.3 on 18,
* 1.4 on 19
* 1.5 on 20

that way we test all dirvers on all elixir versions, we test all otp releases, the only thing that might bite us is a bug that just occurs with elixir 1.5 and erlang 18 for instance which I think is the least likely to occur.

What do you think?

I hope my changes create the build matrix I imagine :D

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,26 @@
 language: elixir
+elixir:
+  - 1.4.5
+  - 1.3.4
+  - 1.5.0
+otp_release:
+  - 19.3
+  - 18.3
+  - 20.0
 matrix:
-  include:
+  exclude:
     - elixir: 1.3.4
+      otp_release: 19.3
+    - elixir: 1.3.4
+      otp_release: 20.0
+    - elixir: 1.4.5
       otp_release: 18.3
     - elixir: 1.4.5
-      otp_release: 19.3
-    - elixir: 1.5.0
       otp_release: 20.0
+    - elixir: 1.5.0
+      otp_release: 18.3
+    - elixir: 1.5.0
+      otp_release: 19.3
 env:
   - ""
   - WALLABY_DRIVER=phantom
--- a/.travis.yml
+++ b/.travis.yml
@@ -54,7 +54,7 @@ before_script:
 
 script:
   - mix coveralls.travis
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -33,24 +35,17 @@ before_script:
   - sudo apt-get install google-chrome-beta
   - sudo apt-get install libstdc++6-4.7-dev
   - MIX_ENV=test mix compile --warnings-as-errors
+  - if [ -z ${WALLABY_DRIVER} ]; then travis_wait mix dialyzer --plt; fi # only run dialyzer for the env without drivers
```
PragTob: @keathley you seem to have removed this line in df1210215c7 - any particular reason or just an accident? :)

PragTob:
So this doesn't create the build matrix I envisioned at all... I thought it'd automatically add all env thingies. Well, onto the next.

PragTob:
So, that seems to do what I want it to do and we'd be down to 15 builds which is still a bit much, but much much better.

.travis.yml
keathley: uh...I'm guessing accident 😬

keathley:
I'm good with just testing each version with its dedicated OTP release. I'm also wondering how likely it is that we could ditch 1.3 support soon.

PragTob:
All passed, we agree this is good --> merge :)

keathley:
LGTM :+1:

PragTob:
Ha forgot to hit the button :see_no_evil:

PR #296, opened 2017-08-21
Prevents failing if phantomjs is not on PATH, but configured manually

### LINKED ISSUE #295: Fails if phantomjs is not installed globally
I have this in my config, as specified in the docs:

```
config :wallabe, phantomjs: "./assets/node_modules/.bin/phantomjs"
```

However, the app still fails until I globally install phantomjs as well (`npm install phantomjs -g`)
Looking at the internal code, it seems that my custom path is still used, but the `Wallabe.Phantom` crashes if a global one does not exist, regardless of the configuration

Is this intended for some reason, or an oversight?

Changing this behaviour in `Wallabe.Phantom.validate()` seems to fix the issue, and my tests still work with only a local installation and not a global one, so I'd guess this is an oversight

(I'll probably open a pull request soon with a fix for this)

naps62:
Fixes #295

### DIFF
--- a/lib/wallaby/phantom.ex
+++ b/lib/wallaby/phantom.ex
@@ -45,13 +45,18 @@ defmodule Wallaby.Phantom do
   end
 
   def validate() do
-    case System.find_executable("phantomjs") do
-      phantom when not is_nil(phantom) ->
+    cond do
+      Application.get_env(:wallaby, :phantomjs, "phantomjs")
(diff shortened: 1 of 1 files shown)


naps62:
Squashed

I also tested this manually on the app I'm currently working on. Not sure if there are any automated tests for this kind of thing (couldn't find any)

Issue #278, opened 2017-07-17
Capture JS logs in chrome


keathley:
It would be good to capture js errors and logs in chromedriver the same way that we do for phantom. I believe we should be able to do this from the chromedriver verbose logs. If not then we'll have to find a way to capture them somehow.

VEkh:
+1 👍

QwertyJack:
I wonder how to hide JS log for phantomjs.

keathley:
@QwertyJack You can change your config: https://github.com/keathley/wallaby#javascript-logging-and-errors

QwertyJack:
Get it, thx.

PR #300, opened 2017-09-11
Capture JS logs in Chrome

chriserin:
We've been using Wallaby at Hashrocket on our [tilex](https://github.com/hashrocket/tilex) project since 0.16 and love it.  We've been using it on client projects as well and its been called at our office "The best integration testing experience so far" by grizzled Rails veterans!  When switching from PhantomJS to Chromedriver --headless the experience was definitely missing the JS errors output and so [Brian Dunn](https://github.com/briandunn) did the coding to add it and I packaged it up into a PR.  We're not especially confident that everything is abstracted tastefully so let us know what you think!

Issue #278

This commit captures js logs for chromedriver in the same way that logs
were captured for phantom.  That is, after each action the driver checks
for logs, appends to the LogStore and parses the logs to be output to
the terminal.

The LogStore and LogChecker have been abstracted for use by all drivers.
Each driver implements its own Logger module.

The chrome driver was delegating to the selenium driver
in many cases, making it hard to wrap the function with a `check_logs!`
call in the same way that the phantom driver did it.  To accomodate
that, we created a private `delegate` function that calls `check_logs`
and the calls `Kernel.apply` to call the appropriate driver function.


PragTob:
Wow, thank you a ton for this and for all the nice words :tada: :rocket: :green_heart: 

Not sure when I'll get to review this one (Although Chris & Aaron might be better suited for this PR either way) but thanks for taking all the time and also for taking the time to put some nice words in there, means a lot :green_heart: 

Enjoy a complimentary picture of cute bunnies!

img_20170731_101343

aaronrenner:
Thanks so much for this PR! 🎉 It looks great at first glance and I'm looking forward to trying it out. I should be able to leave you some feedback in the next couple days.

test/wallaby/experimental/chrome/logger_test.exs
```diff
@@ -0,0 +1,63 @@
...
+      fun = fn ->
+        build_log(~s[http://localhost:52615//logs.html 13:14 Array(2)])
+        |> Logger.parse_log
+      end
+
+      assert capture_io(fun) == "Array(2)\n"
+    end
+
+    test "pretty prints json" do
+      message = ~s(http://localhost:54579//logs.html 13:14 "{\"href\":\"http://localhost:54579//logs.html\",\"ancestorOrigins\":{}}")
```
aaronrenner: I just submitted #301 to fix the issue that's giving you double slashes after the port. Once that's merged, would you mind rebasing and removing these double slashes?

test/wallaby/experimental/chrome/logger_test.exs
```diff
@@ -0,0 +1,63 @@
...
+
+      fun = fn ->
+        "test log"
+        |> build_log()
+        |> Logger.parse_log
+      end
+
+      assert capture_io(fun) == ""
+
+      Application.put_env(:wallaby, :js_logger, :stdio)
```
aaronrenner: #301 Also adds `Wallaby.SettingsTestHelper` which has a `ensure_setting_is_reset/2` that ensures the a setting is reset even if the test fails. Once that PR is merged, would you mind adapting the test to use this helper?

test/wallaby/experimental/chrome/logger_test.exs
chriserin: Yep, will do, I'm watching 301.

test/wallaby/experimental/chrome/logger_test.exs
briandunn: Will do.

lib/wallaby/experimental/chrome.ex
```diff
@@ -87,64 +89,85 @@ defmodule Wallaby.Experimental.Chrome do
...
+  def current_url(session), do: delegate(:current_url, session)
   @doc false
-  defdelegate page_title(session),                                to: WebdriverClient
+  def page_title(session), do: delegate(:page_title, session)
   @doc false
-  defdelegate page_source(session),                               to: WebdriverClient
+  def page_source(session), do: delegate(:page_source, session)
   @doc false
-  defdelegate set_cookie(session, key, value),                    to: WebdriverClient
+  def set_cookie(session, key, value), do: delegate(:set_cookie, session, [key, value])
```
aaronrenner: I know this doesn't affect our public API, but I keep feeling like the `delegate` calls make it difficult to figure out what's going on at a glance. Maybe there's a better name for this function, or maybe it would be worth just writing it out.
```elixir
def set_cookie(session, key, value) do
  check_logs!(session, fn ->
    WebdriverClient.set_cookies(session, key, value)
  end)
end
```
Do you think there is a way to make it more clear what's going on? @keathley, @PragTob what do you think? Is this worth addressing before merging the PR?

keathley: I'm not sure. Writing it out is certainly more explicit. Honestly, I think all of the interactions between the different driver modules and the httpclient are in need of some refactoring. My feeling is that this is alright for now and we can work on refactoring all of these apis as a separate task.  A simple fix for now might be to simple rename `delegate/3` to `webdriver/3` or `httpclient/3`?

aaronrenner: You bring up a good point, @keathley. If we're going to be refactoring the internals of these main driver modules anyways, there's no need worry about the naming of this private function. @chriserin Don't worry about making this change.

chriserin:
Planning to get to this on Friday.

hashrocketeer:
This particular Travis check is failing, WALLABY_DRIVER=selenium WALLABY_SELENIUM_VERSION=3, it fails locally for us too both on this branch and on master --> https://travis-ci.org/keathley/wallaby/jobs/278739707

-@chriserin + @briandunn

aaronrenner:
I just re-ran that job in the build and we're all green now. Sometimes we have to rerun the builds to make it pass.

I wonder if the reason this failed for you locally was because you were running Selenium 3.5 (see #305).

PragTob:
:dancer:

PR #301, opened 2017-09-12
Fixes visit/2 issue where urls have too many slashes

aaronrenner:
I noticed in #300 that `visit/2` is generating urls like http://localhost:54579//logs.html that have too many slashes. This occurs when the base_url has a trailing slash and the requested path has a leading slash. This change adds tests and fixes the issue so we don't end up with urls with too many or not enough slashes depending on how these settings are configured.

### DIFF
--- a/lib/wallaby/browser.ex
+++ b/lib/wallaby/browser.ex
@@ -937,7 +937,10 @@ defmodule Wallaby.Browser do
   end
 
   defp request_url(path) do
-    base_url() <> path
+    base_url = String.trim_trailing(base_url(), "/")
+    path = String.trim_leading(path, "/")
+
+    "#{base_url}/#{path}"
   end
 
   defp base_url do
--- a/test/wallaby/browser_test.exs
+++ b/test/wallaby/browser_test.exs
@@ -1,7 +1,122 @@
 defmodule Wallaby.BrowserTest do
-  use ExUnit.Case, async: true
+  use ExUnit.Case
+
+  import Wallaby.SettingsTestHelpers
 
   alias Wallaby.Browser
+  alias Wallaby.Session
+
+  defmodule TestDriver do
+
+    def start_link() do
+      Agent.start_link(fn -> [] end, name: __MODULE__)
+    end
+
+    def visit(%Session{} = session, path) do
+      Agent.update(__MODULE__, fn (visits) ->
+        [path | visits]
+      end)
+      session
+    end
+
+    def assert_visited(url) do
+      unless Enum.member?(visits(), url) do
+        raise ExUnit.AssertionError,
+          """
+          #{inspect url} was not visited.
+
+          Visited urls:
+          #{visits() |> Enum.map(&"  #{inspect &1}") |> Enum.join("\n")}
+          """
+      end
+    end
+
+    defp visits do
+      Agent.get(__MODULE__, fn (visits) -> visits end)
+    end
+  end
+
+  describe "visit/2" do
+    setup do
+      ensure_setting_is_reset(:wallaby, :base_url)
+      {:ok, _} = TestDriver.start_link()
+      :ok
+    end
+
+    test "relative path without leading slash, base url with trailing slash" do
+      Application.put_env(:wallaby, :base_url, "http://example.com/")
+
+      session = session_for_driver(TestDriver)
+      Browser.visit(session, "form.html")
+
+      TestDriver.assert_visited("http://example.com/form.html")
+    end
+
+    test "relative path with leading slash, base url no trailing slash" do
+      Application.put_env(:wallaby, :base_url, "http://example.com")
+
+      session = session_for_driver(TestDriver)
+      Browser.visit(session, "/form.html")
+
+      TestDriver.assert_visited("http://example.com/form.html")
+    end
+
+    test "relative path without leading slash, base url no trailing slash" do
+      Application.put_env(:wallaby, :base_url, "http://example.com")
+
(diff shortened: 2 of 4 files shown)


lib/wallaby/browser.ex
```diff
@@ -937,7 +937,10 @@ defmodule Wallaby.Browser do
   end
 
   defp request_url(path) do
-    base_url() <> path
+    base_url = base_url() |> String.trim_trailing("/")
+    path = path |> String.trim_leading("/")
```
PragTob: I think style wise piping when there's only one function call in the pipe is discouraged aka maybe rather `String.trim_leading(path, "/")` ?

test/wallaby/browser_test.exs
```diff
@@ -1,7 +1,100 @@
...
+
+  defmodule EchoingDriver do
+    def start_link(test_pid) do
+      Agent.start_link(fn -> test_pid end, name: __MODULE__)
+    end
+
+    def visit(%Session{}, path) do
+      Agent.get(__MODULE__, fn (test_pid) ->
+        send test_pid, {:visit, path}
+      end)
```
PragTob: :+1:

test/wallaby/browser_test.exs
```diff
@@ -1,7 +1,100 @@
...
+      session = %Session{driver: EchoingDriver}
+      Browser.visit(session, "form.html")
+
+      assert_received {:visit, "http://example.com/form.html"}
+    end
+
+    test "relative path with leading slash, base url no trailing slash" do
+      Application.put_env(:wallaby, :base_url, "http://example.com")
+
+      session = %Session{driver: EchoingDriver}
```
PragTob: couldn't we also create the session in the setup and pass it into the tests? Seems to be needed everywhere, no biggy though :)

test/wallaby/browser_test.exs
```diff
@@ -1,7 +1,100 @@
...
+
+    test "relative url when the base_url isn't configured" do
+      Application.delete_env(:wallaby, :base_url)
+
+      assert_raise Wallaby.NoBaseUrl, fn ->
+        Browser.visit(%Session{driver: EchoingDriver}, "/form.html")
+      end
+    end
+
+    test "absolute url when the base_url isn't configured" do
```
PragTob: good test :+1:

lib/wallaby/browser.ex
aaronrenner: Good call. I'll update the code.

aaronrenner:
@PragTob I just updated `Browser.request_path/1` to not use pipes like you suggested. I also renamed `EchoingDriver` to `TestDriver` due to some further refactoring I did on the tests.

I thought about creating a session in the setup block, but then it wouldn't be immediately apparent that we're using a test driver instead of one of the real drivers. I agree defining the struct inline is not very pretty, so I extracted it into a helper.

```elixir
session = session_for_driver(TestDriver)
```

After reading the tests again this morning, when I saw `assert_received {:visit, url}` it wasn't immediately clear why the test process is receiving this message. To figure it out, I would have to go to the old `EchoingDriver` to see this message actually being sent.

Instead of having an `EchoingDriver` that sent the test process messages, I decided to try a custom assertion instead. Also, since the driver wasn't echoing messages anymore, I renamed it to `TestDriver`:

```elixir
TestDriver.assert_visited(url)
```

Now, the test driver keeps a list of all of the urls we've visited in this session and is able to raise a custom `ExUnit.AssertionError` when we don't visit the expected url.

Please let me know what you think and if this approach is clearer than what I had before.

test/wallaby/browser_test.exs
```diff
@@ -1,7 +1,122 @@
...
+
+  import Wallaby.SettingsTestHelpers
 
   alias Wallaby.Browser
+  alias Wallaby.Session
+
+  defmodule TestDriver do
+
+    def start_link() do
+      Agent.start_link(fn -> [] end, name: __MODULE__)
```
PragTob: question about this, we register it under the current module name - so this disallows us from running these test async, correct? Couldn't we just start one and embed the information of the pid of the agent in the `TestDriver`/Session struct so that we could still run them in parallel and not rely on the name?

Alternatively, we could rely on module name + current test description (describe + test) if we can get it, should also be unique.

test/wallaby/browser_test.exs
```diff
@@ -1,7 +1,122 @@
...
+      :ok
+    end
+
+    test "relative path without leading slash, base url with trailing slash" do
+      Application.put_env(:wallaby, :base_url, "http://example.com/")
+
+      session = session_for_driver(TestDriver)
+      Browser.visit(session, "form.html")
+
+      TestDriver.assert_visited("http://example.com/form.html")
```
PragTob: I like putting the assertion in the driver, no implementation leakage that way! :+1:

test/wallaby/browser_test.exs
aaronrenner: Unfortunately we can't run this test case with `async: true` because we're modifying the environment with `Application.put_env(:wallaby, :base_url, "http://example.com")` in some of the tests. 

If that wasn't the case, I agree it would be a great idea to see if we could make this driver not rely on a single registered name. If we reuse this `TestDriver` pattern somewhere else, I'd love to get it to work async.

PragTob: Ah damn, totally I knew I probably overlooked something :D I mean, unless we start putting the base_url in the session right at startup or something like that, but that's a discussion for another time

PR #304, opened 2017-09-13
Implement linting with Credo

### LINKED ISSUE #255: Set up linting
We should really set up linting e.g. [credo](https://github.com/rrrene/credo/) with some sensible configuration for the project.

I believe it makes contributions easier as contributors don't have to fix style problems, which also makes reviews easier and also it makes the whole database more consistent.

tmock12:
Fixes #255

This PR adds linting with Credo and runs `mix credo` in Travis.

I've broken each linting error and its fixes into individual commits so that they could be reviewed commit by commit and any of them could be easily reverted if we decide to ignore that style suggestion. Most of them made sense to me.

A few notes:

- The .credo.exs is ALMOST default:
  - I changed the MaxLineLength to `100` from `80` (033a873). This seemed more reasonable for the time being.
  - I added `include_doc: false` to the TODO specification (8b888a1). Without that our example functions like create_todo/2 in `Wallaby.Browser @module_doc` were showing up as suggestions. 🤷‍♂️
  - I disabled the Consistency check for ExceptionNames (dbfc1d5). I've actually done the name changes but I wanted to put them in a separate PR because they have potential breaking changes. It probably deservers a separate discussion.


PragTob:
Helloooo!

Thank you very much for your contribution :tada: :rocket: 

I been wanting to add credo for a long long time, so this is definitely great. Might take me some time to go through them but just wanted to say thanks and we'll be on it :)

Thanks!
Tobi

keathley:
I'd like to get this merged after #300 is merged. There may be a few extra things that need to be fixed once #300 is in but I think its the cleanest way to do this at the moment. Does that seem alright @tmock12?

keathley:
Just some other thoughts I had about your other notes @tmock12:

* In regards to the issue with `create_todo`, I wonder if thats not a bug in credo. It seems like it should be able to tell the difference between a string todo and a function todo. Perhaps its getting tripped up because its inside the docs? Either way I think we'll be ok with that configuration change.

* I think its reasonable to have a separate PR for the exception name changes. I haven't seen a case in the wild where someone is catching a wallaby exception. Someone might be doing it but we're also pre-1.0 so for now I'm OK with making that change.

tmock12:
Thanks @keathley  

Waiting on #300: makes perfect sense to me. I'll keep an eye on #300 and make changes as necessary after that gets merged in. 

Bug in Credo: I agree that should be a bug in Credo. Its on my list to do a little more investigation around that this week and open up an issue in Credo for at least some discussion about what should and should not show up. Their regexp (https://github.com/rrrene/credo/blob/master/lib/credo/check/design/tag_helper.ex#L12) needs some fine tuning I think. But thats a discussion to have on that Repo I guess 😄 

Separate PR for Exception names: 👍. I'll open it as soon as we get #300 and this one caught up. I've basically got it ready to go but want to give us a chance to get these in first.

lib/wallaby/httpclient.ex
```diff
@@ -1,10 +1,13 @@
 defmodule Wallaby.HTTPClient do
+  @moduledoc """
+  Wallaby HTTPClient that handles making requests to the webdriver API
+  """
```
aaronrenner: This is actually an internal module that's not intended to be in our API docs. Would you mind changing this to `@moduledoc false`?

tmock12: 👍  done

.travis.yml
```diff
@@ -55,6 +55,7 @@ before_script:
 script:
   - mix coveralls.travis
   - if [ -z ${WALLABY_DRIVER} ]; then mix dialyzer --halt-exit-status; fi # only run dialyzer for the env without drivers
+  - mix credo
```
PragTob: Maybe `--strict` ? I mean we can add that later, just what I do usually :)

lib/wallaby/browser.ex
```diff
@@ -516,13 +515,12 @@ defmodule Wallaby.Browser do
     |> find(query)
     |> has_value?(value)
   end
-  def has_value?(%Element{}=element, value) do
+  def has_value?(%Element{} = element, value) do
```
PragTob: Love all of this being fixed up :heart_eyes: 

Thank you so much!

PragTob:
@tmock12 #300 is in so I'm very much looking forward to getting this one :)

tmock12:
Rebased master and should be good to go. #300 changed very little (awesome!). 

Per @PragTob suggestion I added `--strict` to the `mix credo` command. Only thing that changed was a suggestion we alias module names where appropriate, so I did that 5be2a12fa58b3b8fc1a29f00d0b801c8c67e6043

Let me know if anyone has any other thoughts, suggestions or concerns! Thanks for all the work you guys do!

PR #307, opened 2017-09-19
add max_try

QwertyJack:


lib/wallaby/httpclient.ex
```diff
@@ -23,18 +23,22 @@ defmodule Wallaby.HTTPClient do
...
-  defp make_request(method, url, body, retry_count) do
+  defp make_request(method, url, body, _ \\ {0, 5})
+  defp make_request(_, _, _, {m, m}), do: raise "Wallaby had an internal issue with HTTPoison; all #{m} tries failed."
+  defp make_request(method, url, body, {retry_count, max}) do
     HTTPoison.request(method, url, body, headers(), request_opts())
     |> handle_response
     |> case do
          {:error, :httpoison} ->
-           make_request(method, url, body, retry_count+1)
+           :timer.sleep(500 * (retry_count + 1))
```
keathley: Are you just adding a backoff here so that we don't overload the driver? The backoff of 500ms seems pretty high to me. Have you tried other values to see if a lower value is enough to relieve some of the pressure on the driver?

keathley:
@QwertyJack Can you describe the issue that you were having? I can infer the general problem simply based on the solution but it would be good to understand what you were seeing just so we don't end up solving for the wrong thing.

lib/wallaby/httpclient.ex
QwertyJack: Yeah, but I mean the server side. 500ms works fine for me, lower delay also works, such as 100 or 200.

QwertyJack:
When I use wallaby to stimulate a real user using chrome, it always comes up with "Wallaby had an internal issue with HTTPoison". If HTTPoison fails, I do not intend to have another try immediately, but wait for a few second or so.

QwertyJack:
I wonder why build #869.14 failed while #869.15 success. Anyway I use chrome.

aaronrenner:
@QwertyJack Thanks for the PR. I'm still not sure I understand the root cause of the issue. Would you be able to show us a sample project with the issue this PR fixes?

Also, both the Selenium and Chrome drivers are experimental and we definitely have some work to do to make sure they're ready for prime time. That includes improving error messages 😄. `Wallaby.HTTPClient` is an internal API that we forgot to mark with `@moduledoc false`, and if we were going to expose the retry settings to the user, it would be exposed through `Wallaby.Browser`.

Before continuing further on this PR, let's make sure there isn't an async issue. If you could share some code that illustrates the problem you are having, it would be greatly appreciated.

QwertyJack:
See my [toy crawler](https://github.com/QwertyJack/pubmed).

lib/wallaby/httpclient.ex
```diff
@@ -23,18 +23,22 @@ defmodule Wallaby.HTTPClient do
...
+  end
   def request(method, url, params, _opts) do
     make_request(method, url, Poison.encode!(params))
   end
 
-  defp make_request(method, url, body, retry_count\\0)
-  defp make_request(_, _, _, 5), do: raise "Wallaby had an internal issue with HTTPoison"
-  defp make_request(method, url, body, retry_count) do
+  defp make_request(method, url, body, _ \\ {0, 5})
+  defp make_request(_, _, _, {m, m}), do: raise "Wallaby had an internal issue with HTTPoison; all #{m} tries failed."
```
PragTob: `m` doesn't really speak, how I understand it this is `max_tries` or `tries` ?

lib/wallaby/httpclient.ex
```diff
@@ -23,18 +23,22 @@ defmodule Wallaby.HTTPClient do
   def request(method, url, params, [{:encode_json, false} | _]) do
     make_request(method, url, params)
   end
+  def request(method, url, params, [{:max_try, max} | _]) do
+    make_request(method, url, params, {0, max})
+  end
```
PragTob: this adds a capability to add an arbitrary maximum try limit, right? Imo, we don't need that and it's also not called from anywhere, old behaviour should be fine?

lib/wallaby/httpclient.ex
PragTob: Also think that backoff is quite big, I'd also like to see it in a module attribute so that it reads nicer

lib/wallaby/httpclient.ex
QwertyJack: What is a imo?

lib/wallaby/httpclient.ex
QwertyJack: If `tries` equals to `max_tries`, it means all tries have been exhausted but still no success.

lib/wallaby/httpclient.ex
PragTob: An imo is an `in my opinion` relatively common internet jargon, sorry :)

lib/wallaby/httpclient.ex
PragTob: I know what it means :) I'm saying that `m` as a variable name doesn't really communicate that because `m` doesn't say `tries` or `max_tries` but then again, I'm not sure we need this code at all (see bigger review comment)

QwertyJack: I do not get your point; it is a stop condition. Actually I not much familiar with Elixir, but I find this code works fine.

PragTob: the point is not about the code working fine or not, but about readability. Good code is readable and easy to understand sot hat in maintenance, bug fixes and new features it's easier to do

keathley:
Closing this. We're handling a smaller subset of this functionality in #377

PR #308, opened 2017-09-19
Document the Query module, especially specific queries

PragTob:
Feedback very welcome! This is stuff I'd often look up myself by compiling queries while working with wallaby :)

Honestly: I'm not the best at xpath ;P So theoretically I might have gotten some of those wrong.

Some ideas/thoughts:
* Not sure if the attributes should be like they are atm e.g. id or if it should be in ` e.g. `id`
* I had thought about including a doc test that shows a compiled query (I just looked at those when I wrote these) - might help for people, might be too much - my biggest fear is that we'd basically have to adjust these tests every time we change how something is compiled which sounds like a medium big manual churn.

Doc writing style etc. can gladly be commented on, nut sure if all the words are used ideally :)

### DIFF
--- a/lib/wallaby/query.ex
+++ b/lib/wallaby/query.ex
@@ -54,6 +54,20 @@ defmodule Wallaby.Query do
     end
   end
   ```
+
+  ## What does my query do?
+
+  Wanna check out what exactly your query will do? Nothing easier than that with
+  `Wallaby.Query.compile/1` - it takes a query and returns you the css or xpath
+  query it will use:
+
+  ```
+  iex> Wallaby.Query.compile Wallaby.Query.text("my text")
+  {:xpath, ".//*[contains(normalize-space(text()), 'my text')]"}
+  ```
+
+  So, whenever you're not sure whatever a specific query will do just compile
+  it to get all the details!
   """
   alias __MODULE__
   alias Wallaby.Query.XPath
@@ -99,6 +113,9 @@ defmodule Wallaby.Query do
   @type compiled :: {:xpath | :css, String.t}
 
 
+  @doc """
+  Literally queries for the css selector you provide.
+  """
   def css(selector, opts \\ []) do
     %Query{
       method: :css,
@@ -107,6 +124,9 @@ defmodule Wallaby.Query do
     }
   end
 
+  @doc """
+  Literally queries for the xpath selector you provide.
+  """
   def xpath(selector, opts \\ []) do
     %Query{
       method: :xpath,
@@ -115,6 +135,9 @@ defmodule Wallaby.Query do
     }
   end
 
+  @doc """
+  Checks if the provided text is contained anywhere.
+  """
   def text(selector, opts \\ []) do
     %Query{
       method: :text,
@@ -123,6 +146,9 @@ defmodule Wallaby.Query do
     }
   end
 
+  @doc """
+  See `Wallaby.Query.fillable_field/2`.
+  """
   def text_field(selector, opts \\ []) do
     %Query{
       method: :fillable_field,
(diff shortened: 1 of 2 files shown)


lib/wallaby/query.ex
```diff
@@ -168,6 +213,9 @@ defmodule Wallaby.Query do
     }
   end
 
+  @doc """
+  Looks for an option with the given text appears.
```
keathley: I think here we can just say: "Looks for an option that contains the given text".

lib/wallaby/query.ex
```diff
@@ -186,6 +239,10 @@ defmodule Wallaby.Query do
     }
   end
 
+  @doc """
+  Looks a link where the selector is the id, link text, title of the link itself
```
keathley: Smal typo here. Should be "Looks for a link".

lib/wallaby/query.ex
```diff
@@ -195,6 +252,10 @@ defmodule Wallaby.Query do
     }
   end
 
+  @doc """
+  Looke for a file input where the selector is the id or name of the file input
```
keathley: Just a small typo: "Looke" -> "Look"

lib/wallaby/query.ex
```diff
@@ -217,6 +278,18 @@ defmodule Wallaby.Query do
   end
 
   @spec compile(t) :: compiled
+  @doc """
+  Takes a query and compiles it into either css or xpath, which are the queries
+  that will finally be issued to the driver.
```
aaronrenner: How about "Compiles a query into css or xpath so its ready to be sent to the driver"?

lib/wallaby/query.ex
```diff
@@ -54,6 +54,20 @@ defmodule Wallaby.Query do
     end
   end
   ```
+
+  ## What does my query do?
+
+  Wanna check out what exactly your query will do? Nothing easier than that with
+  `Wallaby.Query.compile/1` - it takes a query and returns you the css or xpath
```
aaronrenner: Instead of "Nothing easier than that with `Wallaby.Query.compile/1`", how about "Look no further than `Wallaby.Query.compile/1`"? Also maybe "it takes a query and returns ~~you~~ the css or xpath that will be sent to the driver"

lib/wallaby/query.ex
```diff
@@ -217,6 +278,18 @@ defmodule Wallaby.Query do
...
+  @doc """
+  Takes a query and compiles it into either css or xpath, which are the queries
+  that will finally be issued to the driver.
+
+  ```
+  iex> Wallaby.Query.compile Wallaby.Query.text("my text")
+  {:xpath, ".//*[contains(normalize-space(text()), 'my text')]"}
+  iex> Wallaby.Query.compile Wallaby.Query.css("#some-id")
+  {:css, "#some-id"}
+
```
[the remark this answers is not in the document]
PragTob: Noooo! :trollface:

PragTob:
Thank you for the excellent feedback, all incorporated :rocket:

lib/wallaby/query.ex
```diff
@@ -54,6 +54,18 @@ defmodule Wallaby.Query do
...
   ```
+
+  ## What does my query do?
+
+  Wanna check out what exactly your query will do? Look no further than
+  `Wallaby.Query.compile/1` - it takes a query and returns the css or xpath
+  query that will be sent to the driver:
+
+      iex> Wallaby.Query.compile Wallaby.Query.text("my text")
+      {:xpath, ".//*[contains(normalize-space(text()), 'my text')]"}
```
aaronrenner: Its good to know that you have to indent the doctest instead of using ``` before Elixir 1.4.

PragTob: Yeah remembered that bit, was even surprised the other way works now :D Grew accustomed to the indentation

PragTob:
As we are all green we can :shipit:

Issue #224, opened 2017-05-11
Fix zombie phantomjs processes


aaronrenner:
It looks like since #193 was merged, wallaby is not closing phantomjs processes after the BEAM terminates. I think a better solution (that still keeps wallaby compatible with escripts) would be to write a startup script out to /tmp and then execute it. That should prevent us having to use sub-shells which is causing these zombie proceses. I'm definitely open to other ideas if you have them.

keathley:
Lets try the tmp file for now. I spiked on this at one point and was having some issues with the permissions but you might be able to get something working.

vnegrisolo:
@aaronrenner are you using **mac**? I had the same problem as you described and I fixed the same way as described by @brandonparsons [here](https://github.com/keathley/wallaby/issues/174#issuecomment-292799214).

I don't know the reason but I reinstalled `phantomjs` with brew and I don't have zombie phantomjs processes anymore 💯 

```shell
npm uninstall -g phantomjs-prebuilt
npm uninstall -g phantomjs
brew install phantomjs
```

jameslong:
As a Ubuntu 16.04 user with phantomjs-prebuilt 2.1.1 and Wallaby 0.17 I am seeing this issue. Sorry I don't have time to address this myself but it would be great to have a fix.

keathley:
@aaronrenner Check out some of the changes I'm making to `run_command.sh` in my chromedriver branch. We may be able to resolve these issues that way.

jeroenhouben:
Why was this issue closed? I'm still seeing phantomjs processes. And the list grows when I run the tests multiple times.

I'm on OSX high sierra, installed phantomjs through homebrew

aaronrenner:
It should be fixed in 0.19.2. Are you still seeing the issue with the latest release?

jeroenhouben:
Yes I have `0.19.2` in my mix.lock

aaronrenner:
@jeroenhouben I just tested on OSX 10.11.6 using phantomjs installed through homebrew and didn't see any zombie processes left behind. I'm upgrading my wife's computer to High Sierra to see if I can recreate the issue.

jeroenhouben:
ok thanks for investigating! should i upload a anigif or st of my "situation" would that help?

i see lots of phantoms booting, than tests run, then mosts phantoms disappear except for 2 or 3. repeating this i can see the number of phantomjs processes grow larger and larger

aaronrenner:
[...]

PR #309, opened 2017-09-25
Add support for chrome binary option

opsb:


README.md
```diff
@@ -541,6 +541,18 @@ config :wallaby,
   ]
 ```
 
+### Custom Chrome binary
+
+By default chromedriver will find chrome for you but perhaps you want to test against a different version
```
keathley: Can we change the wording of this to just "but if you want to test against a different version..."

keathley:
The test all look good. I think credo is just upset about some whitespace.

aaronrenner:
Thanks for the PR! I completely understand the problem you're trying to solve, but want to do a little research to see if this is the configuration pattern we want to establish across the rest of our drivers. I'm thinking specifically about what happens when a user tries to pass additional default args through this config and how will that work with our wrapper script?

@keathley @PragTob Could we wait a week or two on this PR until I'm back from vacation and have had a chance to submit my PR for #224? By then, I should have a played around in this area enough to have a more informed opinion on how we should expose this config.

README.md
```diff
@@ -541,6 +541,18 @@ config :wallaby,
...
 ```
 
+### Custom Chrome binary
+
+By default chromedriver will find chrome for you but if you want to test against a different version
+you may use this option to point to the other chrome binary.
+
+```elixir
+config :wallaby,
+  chrome: [
```
aaronrenner: What would you think if we used the driver name as the key? Then it would look something like this.

```elixir
config :wallaby, Wallaby.Experimental.Chrome,
  binary: "path/to/google/chrome"
```
I'm thinking it could be a good convention to use the module name for driver specific config going forward. @keathley, @PragTob Thoughts?

lib/wallaby/experimental/chrome.ex
```diff
@@ -217,6 +218,12 @@ defmodule Wallaby.Experimental.Chrome do
     |> Keyword.get(:headless, true)
   end
 
+  def chrome_binary_option() do
```
aaronrenner: Can this function be private?

README.md
PragTob: :+1: very good idea aaron! :)

lib/wallaby/experimental/chrome.ex
```diff
@@ -232,4 +239,12 @@ defmodule Wallaby.Experimental.Chrome do
       []
     end
   end
+
+  defp put_unless_nil(map, key, value) do
+    if value do
+      Map.put(map, key, value)
+    else
+      map
+    end
```
PragTob: this is totally nitpicking and not sure if others agree, but I'd implement this using 2 methods and pattern matching on nil in the first clause :)

PragTob:
Hey @opsb - wanna fix this up? :)

opsb:
@PragTob I'm not at Zapnito any more so I don't have access to the zapnito fork. I've opened a new PR at https://github.com/keathley/wallaby/pull/355

PR #313, opened 2017-10-05
Allow specifying a name for the screenshot file

nathanl:
I see that there was a some effort to [automatically name screenshots based on the module and test](https://github.com/keathley/wallaby/pull/156), but it's been quiet there for about 7 months.

This PR is simpler: just let users specify the filename, and default to a timestamp as we do now.

### DIFF
--- a/lib/wallaby/browser.ex
+++ b/lib/wallaby/browser.ex
@@ -199,15 +199,17 @@ defmodule Wallaby.Browser do
   Takes a screenshot of the current window.
   Screenshots are saved to a "screenshots" directory in the same directory the
   tests are run in.
+
+  Pass %{name: "some_name"} to specify the file name. Defaults to a timestamp.
   """
   @spec take_screenshot(parent) :: parent
 
-  def take_screenshot(%{driver: driver} = screenshotable) do
+  def take_screenshot(%{driver: driver} = screenshotable, opts \\ %{}) do
     image_data =
       screenshotable
       |> driver.take_screenshot
 
-    path = path_for_screenshot()
+    path = path_for_screenshot(%{name: opts[:name]})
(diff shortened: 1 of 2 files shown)


nathanl:
The test failure for this PR happens only with `WALLABY_DRIVER=selenium WALLABY_SELENIUM_VERSION=3` and also happens on master.

keathley:
I think those test failures are because I've been pre-occupied (some might say lazy) and haven't put together a docker container to make our test environments more stable.

lib/wallaby/browser.ex
```diff
@@ -199,15 +199,17 @@ defmodule Wallaby.Browser do
   Takes a screenshot of the current window.
   Screenshots are saved to a "screenshots" directory in the same directory the
   tests are run in.
+
+  Pass %{name: "some_name"} to specify the file name. Defaults to a timestamp.
   """
   @spec take_screenshot(parent) :: parent
 
-  def take_screenshot(%{driver: driver} = screenshotable) do
+  def take_screenshot(%{driver: driver} = screenshotable, opts \\ %{}) do
```
aaronrenner: I know this is already merged, but what would you think about making opts a keyword list instead of a map and updating the typespec on line 205?

nathanl: @aaronrenner Sounds good to me. Will give it a go.

nathanl:
> I think those test failures are because I've been pre-occupied (some might say lazy) and haven't put together a docker container to make our test environments more stable.

@keathley I meant to come back to this comment and say - you have no reason to feel bad about this whatsoever. Thanks for the free software, and please don't feel guilted into working on it more than you want to. Take a walk instead! 😄

PR #314, opened 2017-10-06
Use a keyword argument list for options here

nathanl:
Basedon @aaronrenner's suggestion on https://github.com/keathley/wallaby/pull/313

@aaronrenner I'm not sure how to update the typespec - feel like helping?

PR #315, opened 2017-10-06
Feature/print screenshot paths for failures

dmcclory:
This is a really small PR that sits on top of #314.

Rspec/Capybara will print out the path to a screenshot when a test fails, which is a tiny small feature I missed here. (eg, in iterm2 you can command-click a the file and open it)

With this PR enabled, output from a test run with failures will look like this:

```
Image Screenshot: file:////Users/dan/prompt/tech_map/screenshots/1507323651592485920.png
Image Screenshot: file:////Users/dan/prompt/tech_map/screenshots/1507323651635410861.png


  1) test cool stuff (ExampleApp.Features.CoolTest)
     test/features/cool_test.exs:10
     ** (Wallaby.ExpectationNotMet) Expected to find 1, visible element that matched the css '.cool-masthead' but 0, visible elements were found.

     code: |> assert_has(css(".cool-masthead"))
     stacktrace:
       test/features/love_cool_test.exs:14: (test)



  2) test organization can be cool (ExampleApp.Features.GreatTest)
     test/features/great_test.exs:10
     ** (Wallaby.ExpectationNotMet) Expected to find 1, visible element that matched the css '.great-masthead' but 0, visible elements were found.

     code: |> assert_has(css(".great-masthead"))
     stacktrace:
       test/features/great_test.exs:14: (test)
```


PragTob:
Hey there!

Thanks for the PR! #314 is now in. This seems like a worthwhile addition. Could you please rebase against master? :)

lib/wallaby/browser.ex
```diff
@@ -212,6 +212,10 @@ defmodule Wallaby.Browser do
     path = path_for_screenshot(name: opts[:name])
     File.write! path, image_data
 
+    if opts[:log_screenshot] do
+      IO.puts "Image Screenshot: file:///#{path}"
```
PragTob: Maybe, text could be adjusted to something like.. "Screenshot taken, find it at #{path}" or something like that.

dmcclory: I just took the text that Capybara uses, but I don't mind changing it. I'll use the text you recommended.

lib/wallaby/browser.ex
```diff
@@ -214,6 +214,10 @@ defmodule Wallaby.Browser do
     path = path_for_screenshot(name)
     File.write! path, image_data
 
+    if opts[:log_screenshot] do
+      IO.puts "Screenshot taken, find it at file:///#{path}"
+    end
```
aaronrenner: What would you think about changing the option name from `:log_screenshot` to `:log`? That way a call with multiple options would look like:

```elixir
take_screenshot(parent, name: 'my_screenshot.jpg', log: true)
```
Also, please update the docs and the typespec for this function. The `take_screenshot_opt` type should be
```elixir
@type take_screenshot_opt :: {:name, String.t} | {:log, boolean}
```

dmcclory:
@aaronrenner - yeah I originally was going to use `:log` and then I made it more complicated for ... some reason. 😅 

docs & typespec are updated as well now.

lib/wallaby/browser.ex
```diff
@@ -201,8 +201,10 @@ defmodule Wallaby.Browser do
   tests are run in.
 
   Pass `[{:name, "some_name"}]` to specify the file name. Defaults to a timestamp.
+  Pass `[{:log, true}]` to log the location of the screenshot to stdout. Defaults to false.
   """
   @type take_screenshot_opt :: {:name, String.t}
+  @type take_screenshot_opt :: {:log, boolean }
```
aaronrenner: I think the two definitions of `take_screenshot_opt` need to be combined like this.

```elixir
@type take_screenshot_opt :: {:name, String.t} | {:log, boolean}
```

keathley:
Thanks for this @dmcclory! My only question is about the formatting. Would it be possible to include the image screenshot paths into the exception text? I think that having those grouped would make it easier to understand which image corresponds to which failure. It seems like that might be a bit more work then whats in this PR to implement so if we want to hold off on that for now we can do that work in a separate PR. What do y'all think @PragTob @aaronrenner?

PragTob:
@keathley good thought, big dan of separate PRs here :+1:

michallepicki:
Moved to https://github.com/keathley/wallaby/pull/436

PR #316, opened 2017-10-09
Make deps examples consistent and test-only

nathanl:
`only: [:test]` is important; otherwise staging or prod may blow up if phantomjs isn't installed

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,7 +71,7 @@ Add Wallaby to your list of dependencies in `mix.exs`:
 
 ```elixir
 def deps do
-  [{:wallaby, "~> 0.19.1"}]
+  [{:wallaby, "~> 0.19.1", only: :test}]
 end
 ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -70,8 +70,8 @@ Read on to see what else Wallaby can do or check out the [Official Documentation
 Add Wallaby to your list of dependencies in `mix.exs`:
 
 ```elixir
-def deps do
-  [{:wallaby, "~> 0.19.1"}]
+def do
```
PragTob: you lost the `deps` somewhere along the way

nathanl: d'oh! That's what I get for editing from the Github UI. 😆

nathanl:
> I think staging/production should just blow up if you do something with wallaby and not by merely including it.

Maybe I did something wrong, but I got this in staging.

crash

PragTob:
hm maybe it's done on application start :|

aaronrenner:
@PragTob That's true, we're starting a pool of PhantomJS workers on application start. I'm planning on fixing that in an upcoming PR.

keathley:
Yeah its the auto-start thing for deps. FWIW I include wallaby in both dev and test. That way I can drive browsers from iex if I just wanna try stuff out.

PR #324, opened 2017-10-14
Update dependencies.

keathley:
Updating our dependencies.


mix.exs
```diff
@@ -49,15 +49,15 @@ defmodule Wallaby.Mixfile do
 
   defp deps do
     [
-      {:httpoison, "~> 0.12"},
+      {:httpoison, "~> 0.13"},
```
PragTob: Is there a feature we need to use from the new httpoison? For runtime requirements I'd always try to keep it as the lowest common denominator, I mean we allow 0.13 already with this :)

keathley: I don't believe there's anything in there that we need. We can keep it at 0.12 for a while longer.

PragTob:
1.4.5 on 19.3 with Selenium 3 is really... fun and blowing up our tests a lot. Restarted build so hopefully it passes now...

keathley:
@PragTob I'm working on a custom docker image that will hopefully make these builds more reproducible (and maybe even faster?).

keathley:
@aaronrenner does this look ok now?

PragTob:
Well then... :shipit:

aaronrenner:
@keathley About the docker image... before you do that, I'd like to make it so the SeleniumDriver handles starting up and monitoring the selenium server like we do with phantom and chromedriver. Also, once we make it so phantom doesn't start up automatically, we can configure our builds to only install webdriver server (phantom/chrome/selenium) that we need. I know it's annoying right now, but if we can make these changes and get things working reliably, we won't have a docker image to maintain.

PR #336, opened 2017-12-24
Add helper assertion `refute_text`

mcasper:
The companion to `assert_text`, this would be a helpful assertion to have when you want to test that text is not on the page for security reasons, or assuring that some sort of transition has happened.

### DIFF
--- a/integration_test/cases/browser/refute_text_test.exs
+++ b/integration_test/cases/browser/refute_text_test.exs
@@ -1,0 +1,30 @@
+defmodule Wallaby.Integration.Browser.RefuteTextTest do
+  use Wallaby.Integration.SessionCase, async: true
+
+  alias Wallaby.ExpectationNotMetError
+
+  test "refute_text/2 waits for presence of text and returns true if not found", %{session: session} do
+    element = session
+              |> visit("wait.html")
+              |> find(Query.css("#container"))
+
+    assert refute_text(element, "not-present")
+  end
+
+  test "refute_text/2 will raise an exception when the text is found", %{session: session} do
+    element = session
+              |> visit("wait.html")
+              |> find(Query.css("#container"))
+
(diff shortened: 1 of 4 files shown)


lib/wallaby/browser.ex
```diff
@@ -588,6 +588,21 @@ defmodule Wallaby.Browser do
     has_text?(parent, text) || raise ExpectationNotMetError, "Text '#{text}' was not found."
   end
 
+  @doc """
+  Matches the Element's content with the provided text and raises if the text is found
+  """
+  @spec refute_text(Element.t, String.t) :: boolean()
+  @spec refute_text(parent, Query.t, String.t) :: boolean()
+
+  def refute_text(parent, query, text) when is_binary(text) do
```
keathley: It doesn't look like this version of the function is tested.

integration_test/cases/browser/refute_text_test.exs
```diff
@@ -0,0 +1,30 @@
...
+  use Wallaby.Integration.SessionCase, async: true
+
+  alias Wallaby.ExpectationNotMetError
+
+  test "refute_text/2 waits for presence of text and returns true if not found", %{session: session} do
+    element = session
+              |> visit("wait.html")
+              |> find(Query.css("#container"))
+
+    assert refute_text(element, "not-present")
```
keathley: I'm not sure this test is really testing that it blocks properly. In order to validate this we'd need to start with some text on the screen and then ensure that refute_text blocks until that text is removed. As it is currently I don't think this will block correctly.

mcasper:
Hey @keathley! The query engine direction makes much more sense, thanks for explaining that. If that's the direction we want to move, I'd be happy to open a PR that trims down the parts of the Browser API that are currently expressible through the query engine (at least `assert_text`)

keathley:
@mcasper If you'd like to work on removing `assert_text` from `Browser` that would be great :+1:

PR #338, opened 2018-01-04
Fixes zombie phantomjs processes

aaronrenner:
After running wallaby as part of a separate application, I confirmed we were still leaving around zombie phantom processes (see #224). It turns out if we kill the script's process group, it will kill all of the child processes as well. This appears to fix the zombie issue once and for all.

One thing to note, I kill the process group by using the negative value of the wrapper script's pid, instead of using `ps axo pid, pgid` to look up the process group id. This is because I was occasionally running into issues where the process group lookup wasn't returning a result, causing
more zombie processes. Hopefully this approach won't cause any issues.

I've tested this on OS X and ubuntu and don't see any processes being left alive. I'm definitely not a shell scripting master, so any ideas for improvement would be greatly appreciated.

### DIFF
--- a/priv/run_phantom.sh
+++ b/priv/run_phantom.sh
@@ -9,6 +9,8 @@
 
 set -e
 
+script_status="running"
+
 create_pipe(){
   local pipe=$(mktemp -u)
   mkfifo -m 600 "$pipe"
@@ -39,10 +41,13 @@ shutdown(){
   local my_pid=$1
   local program_pid=$2
 
-  children=$(ps xao pid,pgid | grep $my_pid | awk '{print $1}' | grep -v $my_pid)
-  kill $program_pid 2>/dev/null
+  if [ $script_status = "running" ]; then
+    script_status="shutting down"
 
-  wait_for_pids_to_exit $children
+    # Kill this script's process group
+    kill -TERM $(($my_pid * -1)) 2>/dev/null
+  fi
+
   exit 0
 }
 


sanderhahn:
We also experienced zombie phantomjs processes, adding `on_exit fn -> Wallaby.end_session(session) end` in our `FeatureCase` setup block seemed to fix this. Not sure if this call is the correct way to fix it, because it is not documented in the readme.

```
    {:ok, session} = Wallaby.start_session(metadata: metadata)
    on_exit fn -> Wallaby.end_session(session) end
    {:ok, session: session}
```

PragTob:
@sanderhahn hmm interesting... I thought sessions should be ended correctly but I'd have to check again

keathley:
@sanderhahn Wallaby monitors each test process (the erlang process the test runs in) and closes the session when it gets a down signal. The code is here: https://github.com/keathley/wallaby/blob/master/lib/wallaby/session_store.ex

Adding the `on_exit` shouldn't cause an issue but if automatic session cleanup isn't working then its probably a separate bug.

PragTob:
restarted the failing job... I think I gotta port over my wrapper from benchee finally...

```
Randomized with seed 369729
** (ExCoveralls.ReportUploadError) Failed to upload the report to 'https://coveralls.io' (reason: timeout).
    lib/excoveralls/poster.ex:20: ExCoveralls.Poster.execute/2
    (mix) lib/mix/tasks/test.ex:245: Mix.Tasks.Test.run/1
    (mix) lib/mix/task.ex:300: Mix.Task.run_task/3
    lib/mix/tasks.ex:51: Mix.Tasks.Coveralls.do_run/2
    (mix) lib/mix/task.ex:300: Mix.Task.run_task/3
    (mix) lib/mix/cli.ex:58: Mix.CLI.run_task/2
    (elixir) lib/code.ex:370: Code.require_file/2
The command "mix coveralls.travis" exited with 1.
0.00s$ if [ -z ${WALLABY_DRIVER} ]; then mix dialyzer --halt-exit-status; fi
```

sparta-developers:
FYI these changes fix the zombie phantomjs problem for us, as well. We just noticed that we had >2000 phantomjs processes on one of our workstations. Yikes!

keathley:
I'll merge this and push out a new point release.

tmecklem:
Sorry to bump this, but I'm encountering this too. Can I somehow entice you to do a release with this fix, @keathley?

keathley:
@tmecklem Sorry about that. Version 0.20 has just been pushed.

PR #343, opened 2018-01-12
Improves readability of file_test failures.

marcdel:
Rather than a failed assertion reporting `Expected truthy, got false`,
it will instead output the values on the left and right hand side making
it easier to diagnose the problem. This isn't a big deal for these
specific tests, but will hopefully make life easier for anyone copying
them into their own project.


PragTob:
Failures are travis infrastructure related:

> The command "sudo apt-get update" failed and exited with 100 during .

Just restarted all of them

PragTob:
sudo problems, restarted all of them...

PR #347, opened 2018-01-24
Always use session url for screenshots

keathley:
We can pass elements or sessions to `take_screenshot`. But using an element url doesn't work in chromedriver. We need to always use the session url even if elements are being passed in.


aaronrenner:
Should be able to close this when #364 is merged.

keathley:
@aaronrenner good call. I'll just do that now :)

Issue #346, opened 2018-01-24
fill_in with emoji value doesn't fill in the input.


sergiotapia:
Screenshot of failing test: 

![image](https://user-images.githubusercontent.com/686715/35313568-1fa73170-008f-11e8-8f01-f84adcc0bf77.png)

Code:

```
|> fill_in(css("input[placeholder='First Name']"), with: "Sergio 🚀")
|> fill_in(css("input[placeholder='Last Name']"), with: "Tapia 🚀")
```

How can I fill in an input with an emoji? I'm trying to write a test for our validation that prevents emojis from being in the first name and last name fields.

Thanks!

keathley:
@sergiotapia which browser are you using? Chromedriver doesn't handle emoji correctly. We might need to figure out a work around for that in the same way gitlab did: https://about.gitlab.com/2017/12/19/moving-to-headless-chrome/. The TL;DR is that they created a javascript function to fill in the text and fire off DOM events.

sergiotapia:
I'm using chromedriver. I'll take a look at Gitlabs workaround and see how I can apply it to our project, then generalize it for Wallaby.

keathley:
I think the solution here is to add something to the README or wiki about this issue. The correct fix should be upstream in chromedriver. We already have an open PR on this here #348 so I think we can close this and manage the fix there.

PR #348, opened 2018-01-24
Fill in with Emoji test

keathley:
## DO NOT MERGE

This reproduces the issue in #346. I'm not sure exactly how we should fix this yet and wanted to get some insight. [Gitlab experienced this same issue](https://about.gitlab.com/2017/12/19/moving-to-headless-chrome) and worked around it by filling in the field with js. Should we consider going that same route?


keathley:
Well, we have to actual do the ~~annoying~~ fun work of fixing it 😝.

PragTob:
easy peasy... ;P

mhanberg:
This appears to be a [chromedriver problem](https://bugs.chromium.org/p/chromedriver/issues/detail?id=2269) and not a Wallaby problem.

I propose that the solution should be to document the behavior and potential workarounds, but not implement any special cases.

@michallepicki What do you think?

michallepicki:
@mhanberg It's awesome that you found the issue tracked in chromedriver!

I agree with @keathley 's comment from here: https://github.com/keathley/wallaby/issues/346#issuecomment-404176302

I think we should document this behavior, link to the chromedriver issue and close this PR. But I think this should be documented here instead of the README, we should clean up the README instead of adding things to it: https://hexdocs.pm/wallaby/Wallaby.Browser.html#fill_in/3

PR #362, opened 2018-04-09
Remove old elixir releases from travis config

keathley:
As the title says, removes old elixir and erlang versions from our travis config.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,26 +1,10 @@
 language: elixir
 elixir:
   - 1.4.5
-  - 1.3.4
-  - 1.5.2
+  - 1.5.3
+  - 1.6.4
 otp_release:
-  - 19.3
-  - 18.3
-  - 20.0
-matrix:
-  exclude:
-    - elixir: 1.3.4
-      otp_release: 19.3
-    - elixir: 1.3.4
-      otp_release: 20.0
-    - elixir: 1.4.5
-      otp_release: 18.3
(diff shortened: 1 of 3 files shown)


PragTob:
:wave: 

Why do we want to do this? Do we want to drop support or do we want to reduce the number of flakies?

I'm just asking because it's rather easy these days to start using methods not available in elixir 1.3/1.4 or some odd API discepreancies to old Erlang versions :)

keathley:
@PragTob Great question :). I really want to remove the flakiness from our test suite. Maybe this isn't a great way to do this. That said I *do* think we should drop support for 1.3.2 at this point but that's a side conversation.

PragTob:
Agreed about dropping 1.3. 1.4 is more... debatable to remove and I think this just reduces the falkyness (fewer runs) but doesn't get rid of it - damn source of it!

So um... leave 1.4 around (just checked - it should be compatible with OTP 20)? Not sure. Whatever we decide to do, we should also adjust the elixir version in mix.exs

.travis.yml
```diff
@@ -1,26 +1,9 @@
 language: elixir
 elixir:
-  - 1.4.5
-  - 1.3.4
   - 1.5.2
```
PragTob: could be upgraded to 1.5.3 btw :)

keathley:
@PragTob I updated credo and fixed the versions :+1:

PR #364, opened 2018-04-15
Finish chromedriver support.

keathley:
This PR does a few things. I would have normally broken this up but I wanted to get all of our chrome issues sorted at one time and I just got carried away. Here's the highlights:

* Fixes an issue with taking screenshots on chrome. Chrome doesn't support screenshots based on elements so we need to always screenshot using the session.
* Captured a specific chromedriver error for invalid selectors
* Fixes the dialog support for chrome. I'll discuss this in more detail below.
* Enable all of the tests for chrome.

In order to support dialogs and alerts in chrome I made the decision to remove `accept_dialogs` and `dismiss dialogs`. These were useful functions in phantom because phantom has no good way of managing dialog boxes. With chrome and selenium these functions are less useful since we don't have the limitations of phantomjs. I also felt that having these functions were simply less useful then simply being explicit by using `accept_prompt`, `accept_alert`, and the others. Thinking through it I felt like it would be good to remove a bit of the maintenance burden by deleting those functions.

I know its a lot of stuff but let me know what you think.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,6 +47,7 @@ before_script:
   - sudo add-apt-repository ppa:ubuntu-toolchain-r/test --yes
   - sudo apt-get update
   - sudo apt-get install google-chrome-beta
+  - sudo apt-get --only-upgrade install google-chrome-stable
   - sudo apt-get install libstdc++6-4.7-dev
   - MIX_ENV=test mix compile --warnings-as-errors
   - if [ -z ${WALLABY_DRIVER} ]; then travis_wait mix dialyzer --plt; fi # only run dialyzer for the env without drivers
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,7 +46,6 @@ before_script:
   - sudo mv chromedriver /usr/local/bin
   - sudo add-apt-repository ppa:ubuntu-toolchain-r/test --yes
   - sudo apt-get update
-  - sudo apt-get install google-chrome-beta
   - sudo apt-get --only-upgrade install google-chrome-stable
   - sudo apt-get install libstdc++6-4.7-dev
   - MIX_ENV=test mix compile --warnings-as-errors
--- a/integration_test/chrome/all_test.exs
+++ b/integration_test/chrome/all_test.exs
@@ -1,32 +1,7 @@
-# Code.require_file "../tests.exs", __DIR__
-Code.require_file "../cases/browser_test.exs", __DIR__
-Code.require_file "../cases/browser/all_test.exs", __DIR__
-Code.require_file "../cases/browser/assert_css_test.exs", __DIR__
-Code.require_file "../cases/browser/assert_refute_has_test.exs", __DIR__
-Code.require_file "../cases/browser/assert_text_test.exs", __DIR__
-Code.require_file "../cases/browser/attr_test.exs", __DIR__
-Code.require_file "../cases/browser/clear_test.exs", __DIR__
-Code.require_file "../cases/browser/click_button_test.exs", __DIR__
-Code.require_file "../cases/browser/click_test.exs", __DIR__
-Code.require_file "../cases/browser/cookies_test.exs", __DIR__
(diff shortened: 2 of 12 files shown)


keathley:
Looks like our chrome version needs to be over 64.0.3282.

keathley:
Our chrome issues should now be fixed. Once this is merged in we can rebase some of the other PRs and get those merged as well.

.travis.yml
```diff
@@ -47,6 +47,7 @@ before_script:
   - sudo add-apt-repository ppa:ubuntu-toolchain-r/test --yes
   - sudo apt-get update
   - sudo apt-get install google-chrome-beta
+  - sudo apt-get --only-upgrade install google-chrome-stable
```
aaronrenner: Just out of curiosity, why do we install chrome beta and then upgrade `google-chrome-stable`? Could we skip installing chrome beta, since it appears headless chrome has made it's way into stable?

lib/wallaby/driver.ex
```diff
@@ -30,11 +30,6 @@ defmodule Wallaby.Driver do
   """
   @callback accept_confirm(Session.t, open_dialog_fn) :: {:ok, [String.t]} | {:error, reason}
 
-  @doc """
-  Invoked to accept all open dialogs.
-  """
-  @callback accept_dialogs(Session.t) :: {:ok, String.t} | {:error, reason}
-
```
aaronrenner: We'll need to remember to make a note in the readme about this breaking change

lib/wallaby/experimental/selenium/webdriver_client.ex
```diff
@@ -235,7 +235,7 @@ defmodule Wallaby.Experimental.Selenium.WebdriverClient do
   """
   @spec take_screenshot(Session.t) :: binary
   def take_screenshot(session) do
-    with {:ok, resp}   <- request(:get, "#{session.url}/screenshot"),
+    with {:ok, resp}   <- request(:get, "#{session.session_url}/screenshot"),
```
aaronrenner: It looks like the typespec is incorrect here too. It should probably be
```elixir
@spec take_screenshot(Session.t | Element.t) :: binary
```

I didn't realize we could pass an element to this function, but I checked the tests and appears that we support it. 😄

integration_test/chrome/all_test.exs
```diff
@@ -1,32 +1,7 @@
...
-Code.require_file "../cases/browser/navigation_test.exs", __DIR__
-Code.require_file "../cases/browser/page_source_test.exs", __DIR__
-Code.require_file "../cases/browser/select_test.exs", __DIR__
-Code.require_file "../cases/browser/set_value_test.exs", __DIR__
-Code.require_file "../cases/browser/send_text_test.exs", __DIR__
-Code.require_file "../cases/browser/stale_nodes_test.exs", __DIR__
-Code.require_file "../cases/browser/text_test.exs", __DIR__
-Code.require_file "../cases/browser/title_test.exs", __DIR__
-Code.require_file "../cases/browser/visible_test.exs", __DIR__
+Code.require_file "../tests.exs", __DIR__
```
aaronrenner: Sweet, we support everything now. 🎉 Thanks for doing this!

.travis.yml
keathley: yeah turns out installing chrome beta doesn't help because it wasn't being used anyway. Its left over from when I was trying to figure out how to get an updated chrome installed. I'll remove it.

lib/wallaby/experimental/selenium/webdriver_client.ex
keathley: We allow either to be passed in. With phantom it would actually take a picture of the element. But with the other drivers its a bug.

integration_test/chrome/all_test.exs
keathley: Took long enough haha. Glad its finally coming together.

keathley:
@aaronrenner Should be fixed up :)

PR #369, opened 2018-05-14
upgrade httpoison dependency to ~> 1.1

jaimeiniesta:
Hey there!

The dependency on HTTPoison ~>0.12 is quite old, this PR upgrades it to ~>1.1


mix.exs
```diff
@@ -49,7 +49,7 @@ defmodule Wallaby.Mixfile do
 
   defp deps do
     [
-      {:httpoison, "~> 0.12"},
+      {:httpoison, "~> 1.1"},
```
NobbZ: I think the best for now where to use `"~> 0.12 or ~> 1.0"`, this should work with both while also not breaking the ugly way after a breaking change in HTTPoison 2.0 happens.

NobbZ:
I've provided a comment inline, but can't see it right now, therefore I'll repost:

> I think the best for now where to use `"~> 0.12 or ~> 1.0"`, this should work with both while also not breaking the ugly way after a breaking change in HTTPoison 2.0 happens.

Additionally I think, CI should be set up in a way to check against 0.12 for downwards compatibility, while also running against 1.x to make sure it behaves correctly on recent versions as well.

jaimeiniesta:
Yes, I think a more relaxed dependency like `"~> 0.12 or ~> 1.0"` would be better for more flexibility.

PragTob:
@NobbZ generally yes, I think we'll now test with the newer versions. However, setting it up on CI would likely lead to another explosion in our test matrix which is already quite big :| Not sure how to best handle that

NobbZ:
Yeah, I do see the problem with the build-matrix.

keathley:
We can use the latest in CI for now. If people have an issue with the older version they're likely to upgrade anyway IMO.

Issue #306, opened 2017-09-18
xpath selectors need to use exact text for label fields.


keathley:
This is based on #257. Currently our xpath selectors for label text use `contains(#{actual_text})`. This can lead to "Multiple Element Errors" if 2 different fields have similar label texts. I think it would be good to match against the exact label text for an element instead of using `contains()`. This change would require users to use a fully qualified label text but, anecdotally, thats all I've ever seen anyone do in practice anyway.

zedtux:
@keathley what about this issue? I'm facing exactly the issue as @jeffdeville in my application (`Password` and `Password confirmation`) and I'm looking for a solution.

keathley:
For now you should be able to use the id of the element like `text_field("password")` and `text_field("password_confirmation")`.

zedtux:
Thank you @keathley it is working 👍 but still do you have an idea when this could be addressed?

keathley:
I'll try to prioritize this fix.

PR #387, opened 2018-10-11
Stop using contains for queries which use labels to match

joelbyler:
I'm not sure if this is the perfect fix for #306 but I was able to get the new test to pass using this revised xpath.  That thing is so hard to follow though, I'm sure I might have missed something.  I'm opening this in the hopes to start a conversation about fixing this issue and I'm glad to help out.  If this does actually fix the issue then 🎉, but I'm betting there's more to it.

Also, what are your thoughts on making that xpaths easier to parse visually?  I spent some time breaking bits of it apart locally but didn't commit any of that here.  Perhaps I can offer some help there to lower the bar of entry for others, let me know.  

Thanks for reviewing and thanks in advance for your feedback!


joelbyler:
... and there is a check that labels have a for attribute.  Hmm...  I'll look into that.

keathley:
@joelbyler Thanks for working on this! I'm sorry about the xpath stuff. It's gross. I originally intended on building a library to compose xpath queries but never made time for it. I haven't come up with a better solution yet but if you have thoughts I'd be interested in hearing them.

fusillicode:
Hi guys 😄 

is this still relevant? I would love to align this PR with the latest master if needed 😊

P.S: thanks a lot for this great library! 🙇

PR #388, opened 2018-10-16
Looking to get the discussion started regarding XPath module refactoring

joelbyler:
I'm not sure if this is the right direction but I'm looking for your feedback.  I feel like we could build up a module with more expressions that can be composed together for building the xpath expressions which are appropriate for each form control.  

Also, I've noticed some quirkiness around how each expression works for the different types of form controls, after some consideration and experimentation these seem intentional but makes it difficult to extract larger expressions to be useful across the board, I'm wondering if you have ideas here.

### DIFF
--- a/lib/wallaby/query/xpath.ex
+++ b/lib/wallaby/query/xpath.ex
@@ -2,6 +2,8 @@ defmodule Wallaby.Query.XPath do
   @moduledoc false
   # credo:disable-for-this-file Credo.Check.Readability.MaxLineLength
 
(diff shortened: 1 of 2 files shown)


keathley:
I think this looks like a great start! One way that we might be able to tackle the quirkyness of some of these selectors is to extract "lower-level" building blocks that can be composed together.

Originally I started doing something like this:

```
```

mhanberg:
I'm going to close this, but will definitely revisit it when we look at the Xpath stuff again.

Issue #152, opened 2017-03-05
Responsive design and window size


DavidVII:
After bringing in some css my app now has a responsive navigation where a window size narrower than 769 pixels wide will turn into a hamburger menu. Since the default window size in wallaby is 100 x 100 I end up running tests against the responsive layout and can not click on links in the responsive menu since it involves an extra step. This leaves me with two options:

1. Account for extra step
2. resize_window in the current session

I opted for the second option, but I have quite a few tests where I have to keep resizing the window. Is there a way to configure the default window size for all acceptance cases? I'd love to set this option globally so that I don't have to keep resizing across all my tests. I'm happy to help if this option isn't available. I just wanted to double check a feature like this doesn't already exist.

Cheers!

keathley:
I don't think there's a window size option to pass on session creation in webdriver. You will have to set it after creating the session. You could add something like this to your AcceptanceCase:

```elixir
{:ok, session} = Wallaby.start_session(metadata: metadata)
session = Wallaby.Browser.resize_window(session, 300, 500)
{:ok, session: session}
```

That way each session would be sized correctly. We could probably support this as a configuration option. It probably makes sense to open to window larger then 100x100 by default.

PragTob:
> It probably makes sense to open to window larger then 100x100 by default.

Most definitely. I imagine it being a gotcha moment in the first couple of tests that that's why maybe the menu doesn't show :D

DavidVII:
Thanks a lot for that, @keathley. That clears things up quite a bit. I was wondering how to go about setting that in the `AcceptanceCase` file.

However, I'm wondering now weather or not a config is necessary. Perhaps adding a line or two about that in the README could suffice?

What do you guys think would be a good default? Perhaps something like 960x500? I'm happy to submit a PR over the next few days if you'd like.

PragTob:
I think an addition to the README would be quite nice, and agree that it's an easy enough change that it doesn't need a config imo :)

vanderhoop:
Just got nailed by this myself! I'm testing real-time updates across different sessions, so I'm consistently starting a second session. For now, I'm resizing the second session's window winthin my real-time update tests such that it matches the size I've given to the first session in my `IntegrationCase` setup, but @keathley, I'm happy to submit a PR for passing screensize options to `Wallaby.start_session` if you're down with it. 

Instead of 

```elixir
 {:ok, session_two} = Wallaby.start_session()
 session_two = resize_window(session_two, 960, 500)
```
[...]

PR #390, opened 2018-10-25
Option to configure default window size

mhanberg:
Addresses #152 

@keathley @PragTob @aaronrenner Let me know what you think.

Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -425,13 +425,26 @@ session
 |> assert_has(css(".alert", text: "Welcome!"))
 ```
 
-### Windows and Screenshots
+### Window Size
 
-It's possible to interact with the window and take screenshots:
+You can set the default window size with the `window_size` config option.
+
+```elixir
+config :wallaby, window_size: [width: 1280, height: 720]
+```
+
+You can also resize the window and get the current window size during the test.
 
 ```elixir
 resize_window(session, 100, 100)
 window_size(session)
+```
+
+### Screenshots
+
+It's possible take screenshots:
+
+```elixir
 take_screenshot(session)
 ```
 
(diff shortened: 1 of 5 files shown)


mhanberg:
One test failure is for `travis.coveralls`, the others I'm not so sure about. 

I wasn't able to run all of the selenium tests locally. Most were failing with something like this

```
Cannot find function clause

defp something(%{"element-235h3kj45h3k4h523j45" => "something})

did you mea

defp something(%{"ELEMENT" => "something
```

so I'm not sure if the travis errors are legitimate or not.

keathley:
@mhanberg Thanks! Let me take a look at the test failures.

keathley:
This code looks good to me. @mhanberg I merged your PR for returning the correct errors from http calls. If you want to rebase this branch against master then we can see what the real error is. Its happening on test runs for other branches as well so its either that something is broken on master or broken in our CI or something.

mhanberg:
I rebased and it looks like the same results (except with an extra `travis.coveralls` failure).

The test failures are all with `WALLABY_SELENIUM_VERSION=3`, so I'm guessing it has something to do with that. Is that configured to be a specific version of selenium, or could it have been changed silently? (meaning something like `latest` instead of `1.1.1`)

keathley:
Looks like we're using an explicit version of selenium: https://github.com/keathley/wallaby/blob/master/test/tools/start_webdriver.sh#L15 and ghecko driver but "latest" version of firefox: https://github.com/keathley/wallaby/blob/master/test/tools/start_webdriver.sh#L30.

I suspect that this might be the issue.

keathley:
Just checked and it looks like the most recent version of selenium is 3.14 and most recent version of ghecko driver is 0.23. @mhanberg if you wanna submit a separate PR for those fixes to see if they fix the problem that would be great. Otherwise it'll be a bit before I can get around to it.

mhanberg:
@keathley I can take care of that. Thanks for digging into it :the

mhanberg:
@keathley we're green on Travis now thanks to the other PR's 👍

keathley:
Thanks @mhanberg :tada:

PR #391, opened 2018-10-26
silence quoted keyword warning from recent elixir versions

tmepple:
A minor update to silence the compiler warning `found quoted keyword "coveralls" but the quotes are not required.` coming from a recent version of elixir


tmepple:
Thanks for looking into it... weird there would be test failures for such a simple change to quiet compiler warnings.  That's what I get for being lazy and pushing the commit without running the tests locally... apologies for that.

keathley:
@tmepple I doubt it was your fault. I think something unrelated is broken for the latest selenium 3. No worries 👍.

PR #394, opened 2018-10-30
Don't fail the build on Coveralls timeout

mhanberg:
Coveralls timing out is causing a lot of false positives on CI. This should recover from excoveralls failures.

Shamelessly ripped from [pinterest/elixir-thrift](https://github.com/pinterest/elixir-thrift/pull/389/files)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,7 +37,7 @@ before_script:
   - bash $TRAVIS_BUILD_DIR/test/tools/start_webdriver.sh
 
 script:
-  - mix coveralls.travis
+  - mix coveralls.safe_travis
   - if [ -z ${WALLABY_DRIVER} ]; then mix dialyzer --halt-exit-status; fi # only run dialyzer for the env without drivers
   - mix credo --strict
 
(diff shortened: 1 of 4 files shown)


mhanberg:
@keathley I've got this passing, except for some dialyzer warning that I don't understand (I haven't yet worked with dialyzer in any capacity).

Do you have any ideas?

```
done in 0m4.65s
:0: Unknown function 'Elixir.Mix':shell/0
:0: Unknown function 'Elixir.Mix.Tasks.Coveralls':do_run/2
lib/mix/tasks/safe_travis.ex:1: Callback info about the 'Elixir.Mix.Task' behaviour is not available
```

I have a feeling it has to do with ExCoveralls being a test env dependency, but I'm not sure.

keathley:
I think the issue is that its not building the plt with mix as an included application. You may have to adjust the PLT options in order to add mix. In the meantime you can pretty safely skip those errors by including something like `@dialyzer {:nowarn_function, 'Elixir.Mix:shell': 0}` into that module.

mhanberg:
@keathley I was able to ignore the warnings by adding them to a `dialyzer.ignore_warnings` file.

With that done, I think this is ready for your consideration. You can see that on my [fork](https://travis-ci.com/mhanberg/wallaby/), the tests all pass even though the travis upload fails.

keathley:
Thanks @mhanberg :tada:

Issue #266, opened 2017-06-29
It would be nice to pretty-print a `Wallaby.Element`


knewter:
Right now I can do this if I pipe it and get the `outerHTML`:

```elixir
some_element
|> Element.attr("outerHTML")
```

but it's not so pretty :) Would be neat to get a nice output of "what exactly did I grab here?" for debugging purposes. Only added this issue because @PragTob suggested it, otherwise I'd feel like I was piling on personal wishlist features :)

nathanl:
This issue helped me a lot with debugging.

I had a test with `assert_has(css("some_selector"))` that was failing, and I wanted to figure out why.

I tried using `take_screenshot/1` to see the page, but it appeared that the element I wanted had been rendered.

I tried using `page_source/2` to see the HTML, but that showed me the page before JavaScript did any DOM manipulation, and I was looking for something it would create.

Finally, I stumbled on this issue and figured out I could see the DOM as modified by JavaScript by using:

```elixir
File.write("/tmp/the_page.html", Element.attr(find(session, css("body")), "outerHTML"))
```
Inspecting that output showed me what I was asserting incorrectly.

I'm going to hang on to this trick, but it might be nice to have a more straightforward way to inspect the current DOM.

keathley:
One solution for this would be to use the inspect protocol. I'm not sure how I feel about having side-effects in there or if the results would end up being surprising to folks. Otherwise we could have something like `Element.inspect/1`. What do y'all think?

PragTob:
Definitely no side effects in the inspect protocol please. A separate function to inspect elements is highly preferred to my mind :)

johnmosesman:
@keathley Not sure if this fits within `take_screenshot/1` or not, but it would be great to have an option to save the output as an HTML file. The image works well for visually seeing if the elements are there, but there are many times when I want to inspect the HTML structure.

PragTob:
@johnmosesman I think it should be a different function call, like `dump_html` or whatever. I have a helper function somewhere to at least print it for myself :)

nathanl:
I've added this to my `FeatureCase`:

```elixir
using do
  quote do
    use Wallaby.DSL

    alias Mealthy.Repo
    import Ecto
    import Ecto.Changeset
    import Ecto.Query

    import Mealthy.Factory
    import Mealthy.Web.Router.Helpers

    def dump_html_to(session, filename) do
      File.write(
        filename,
        Wallaby.Element.attr(
[...]

PR #396, opened 2018-11-11
Print outerHTML when passing an Element to IO.inspect/2

mhanberg:
Addresses #266 

When passing a `Wallaby.Element` to `IO.inspect/2`, the elements outerHTML will be printed along with the normal representation of a struct, like so:

image

I would like some feedback on a few things:

- Where is the best place to define the `Inspect` implementation?
- Keep/change/throw out the coloring?
- Is the format alright?
- Should we request the `outerHTML` data when running inspect, or should the outerHTML be added to the `Wallaby.Element` struct itself?

Thanks!

cc/ @keathley @knewter @nathanl

### DIFF
--- a/integration_test/cases/inspect_test.exs
+++ b/integration_test/cases/inspect_test.exs
@@ -1,0 +1,39 @@
+defmodule Wallaby.Integration.InspectTest do
+  use Wallaby.Integration.SessionCase, async: true
+  import ExUnit.CaptureIO
+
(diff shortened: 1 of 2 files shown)


nathanl:
Your code and the results look great to me! 👏 😄 

> Should we request the outerHTML data when running inspect, or should the outerHTML be added to the Wallaby.Element struct itself?

I don't know the best answer to this part.

PR #399, opened 2018-11-29
Allow newlines in chrome log messages

tmepple:
I ran into this issue today where some of the logs returned by chromedriver for one of my tests had embedded newlines which caused the logger Regex to fail (return nil) which led to throwing a `CaseClauseError` here:  https://github.com/keathley/wallaby/blob/bfddea8a7806e6a8787741577ea4054aab846a9c/lib/wallaby/experimental/chrome/logger.ex#L14-L16

This PR adds the `s` Regex modifier allowing newlines to be matched by the dot (.)


keathley:
Thanks @tmepple :tada:

PR #400, opened 2018-12-12
Make phantom start timeout configurable

studzien:
This PR is to start the discussion on a possible way of solving the following issue:

On my and my colleague's Mac, `wallaby` regularly (1 in 10 times, especially when Mac is doing something else) fails to start, because `phantomjs` start times out after 5 seconds.
Increasing this value resolves this.

Is there a specific reason why the `phantomjs` start happens in an asynchronous task?
Is it possible that this process hangs forever and thus the timeout needs to be controlled with `Task.await`?

### DIFF
--- a/lib/wallaby/phantom/server.ex
+++ b/lib/wallaby/phantom/server.ex
@@ -95,8 +95,8 @@ defmodule Wallaby.Phantom.Server do
 
   @spec start_phantom(ServerState.t) ::
     {:ok, ServerState.t} | {:error, StartTask.error_reason}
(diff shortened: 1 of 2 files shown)


keathley:
I honestly don't remember why we needed to start phantom in a task. I feel like it might have had something to do with error handling. Might be worth looking at the history of that file and see why we ended up in that situation. I think this fix is fine but its probably worth revisiting to see if there might be a different design.

Long term its probably best to put effort into better headless chrome support.

Issue #380, opened 2018-07-13
Add selected to query api


keathley:
There is currently no way to specify if a value is "selected" at the query level. This means if you want to assert an element is selected you need to do this:

```elixir
session
|> find(elem, fn e -> assert Element.selected?(e) end)
```

This is error prone since its no longer declarative and has race conditions.

We need to add `selected` to our list of query attributes.

mhanberg:
I'm going to start working on this

germsvel:
I was doing some preliminary research on this, and I had a couple of questions. Sorry for the long questions; let me know if they're not clear, and I can elaborate more. 

---

The first one is just a clarifying question: 

Is the intention of this to create something like this? (just want to confirm before diving in)

```elixir
session
|> find(Query.selected("#option-1"))
```

---

The second question is about the argument being passed into `Query.selected/1` (assuming the first question is correct). I know most of the form elements (e.g. `checkbox`) take a selector which could be a [name, id text, or label text](https://hexdocs.pm/wallaby/Wallaby.Query.html#module-form-elements). Was the intention for this to use xpath behind the scenes or a css selector?

I ask, because apparently xpath is [unable to find an option that is currently selected](https://stackoverflow.com/a/17988465) doing something like this `.//option[@selected='selected']`. I tested that with the `forms.html` fixture, and we can find the default selection, but it will not find the element selected if it has been changed. So it seems we _cannot_ use xpath for `Query.selected/1` (unless there's another way I'm unaware of). 

It does seem like we could use the css selector [:checked](https://developer.mozilla.org/en-US/docs/Web/CSS/:checked), but then we wouldn't be able to search by label text (e.g. `Query.selected("Option 1")`. We could let it be a css selector, and we append `:checked` to it: 

```
# tests
Query.selected("#option-1")

#compiled
{:css, "#option-1:checked"}
```

I'm not sure if it makes sense to go with css selectors or not here. So I'd be curious to hear your thoughts on it.

keathley:
Sorry for the delay @germsvel and thanks for looking into this.

As for the first question I was originally thinking of something like this: `css("#option-1", selected: true)`. That way we can use our existing queries but determine if they are selected or not. As for implementation I think we may have to do something similar to how we check for visibility. The way that works is we find all the elements that match the original selector and then filter them based on the other provided options. There is a `selected` webdriver call that the driver module should already support.

Let me know if you have other questions or if that doesn't make sense.

germsvel:
[...]

PR #402, opened 2019-01-07
Add selected condition to query

germsvel:
This is an attempt to handle https://github.com/keathley/wallaby/issues/380

Description
===========

Adds a `selected` condition to queries. This should only work with options, radio buttons, and checkboxes (since it uses `Element.selected?/1` underneath).

The usage is similar to the `visible` option for a query function: e.g. `Query.css("selector", selected: true)`. The default case, when neither `selected: true` nor `selected: false` is passed sets the condition to `:any`. In this way, if neither `true` nor `false` are specified, the query will return both selected and unselected elements. Having `true`, `false` or `:any` feels a bit unclean to me since now we have a boolean or a random atom `:any`, but I couldn't think of a better way to handle that. I'm happy to change that if there are better options. 

I also updated the error message to include the notion of whether selected or unselected elements were being searched for -> `"... 1, visible, selected element, but ..."`. I wasn't sure if this was desired, so I'm happy to remove this if it doesn't make sense. 

Thanks so much for the time spent reviewing this, and let me know if you'd like any changes or if we should just close this altogether and try a different approach.

### DIFF
--- a/lib/wallaby/query/error_message.ex
+++ b/lib/wallaby/query/error_message.ex
@@ -78,7 +78,7 @@ defmodule Wallaby.Query.ErrorMessage do
 
   defp found_error_message(query) do
     """
-    #{expected_count(query)}, #{visibility(query)} #{method(query)} #{selector(query)} but #{result_count(query.result)}, #{visibility(query)} #{short_method(query.method, Enum.count(query.result))} #{result_expectation(query.result)}.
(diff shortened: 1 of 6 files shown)


lib/wallaby/query/error_message.ex
```diff
@@ -174,6 +174,15 @@ defmodule Wallaby.Query.ErrorMessage do
   end
   def condition(_), do: nil
 
+  @spec visibility_and_selection(Query.t) :: String.t
+  defp visibility_and_selection(query) do
```
germsvel: This is how we're managing the combined error message `"... 1 visible, selected element, but ..."`. I'm happy to change this if there's a better way or to remove it if we don't want to include `selected/unselected` in the error message.

Issue #386, opened 2018-08-16
Session ID is nil after creating session


mschae:
To preface: We are trying to use Wallaby inside a docker container, talking to selenium in a different container.

## Issue

When invoking `Wallaby.start_session/2` it returns `nil` for the session ID.

## Test Code & HTML

`config/test.exs`:
```Elixir
config :wallaby,
  base_url: "http://backend:4001",
  driver: Wallaby.Experimental.Selenium
```

`test/support/integration_test.ex`:
```Elixir
# ...
    {:ok, session} =
      Wallaby.start_session(remote_url: "http://selenium_chrome:4444/wd/hub/")
      |> IO.inspect()
# ...
```

Result:
```
{:ok, 
  %Wallaby.Session{
    driver: Wallaby.Experimental.Selenium,
    id: nil, 
    screenshots: [],
    server: :none,
    session_url: "http://selenium_chrome:4444/wd/hub/s
ession/",
    url: "http://selenium_chrome:4444/wd/hub/session/"
}}
```

I logged the response of the selenium server and it is as follows:
```
%{
  "status" => 33,
  "value" => %{
    "error" => "session not created",
    "message" => "Unable to create session from org.openqa.selenium.remote.NewSessionPayload@3a1376aa\nBuild info: v
ersion: '3.12.0', revision: '7c6e0b3', time: '2018-05-08T15:15:08.936Z'\nSystem info: host: 'b47680a9db23', ip: '192
.168.224.4', os.name: 'Linux', os.arch: 'amd64', os.version: '4.9.93-linuxkit-aufs', java.version: '1.8.0_171'\nDriv
er info: driver.version: unknown",
    "stackTrace" => [
      %{
        "className" => "org.openqa.selenium.remote.server.NewSessionPipeline",
        "fileName" => "NewSessionPipeline.java",
        "lineNumber" => 75,
        "methodName" => "lambda$null$4"
      },
      %{
        "className" => "java.util.Optional",
        "fileName" => "Optional.java",
        "lineNumber" => 290,
        "methodName" => "orElseThrow"
      },
      %{
        "className" => "org.openqa.selenium.remote.server.NewSessionPipeline",
        "fileName" => "NewSessionPipeline.java",
        "lineNumber" => 74,
        "methodName" => "lambda$createNewSession$5"
      },
      %{
        "className" => "java.util.Optional",
        "fileName" => "Optional.java",
        "lineNumber" => 267,
        "methodName" => "orElseGet"
      },
      %{
[...]

PR #408, opened 2019-01-31
When creating session, read sessionID from value object when it's present

michallepicki:
Newer platforms, like the web drivers for Safari 12.0 on macOS Mojave or Edge 18 on Windows 10, seem to respond with a different response format (tested on Browserstack Automate). It may be that this is aligned the new W3C WebDriver spec, but I haven't checked.

Related issue: #386

### DIFF
--- a/lib/wallaby/experimental/selenium.ex
+++ b/lib/wallaby/experimental/selenium.ex
@@ -37,7 +37,7 @@ defmodule Wallaby.Experimental.Selenium do
     capabilities = Map.merge(default_capabilities(), capabilities)
 
     with {:ok, response} <- create_session_fn.(base_url, capabilities) do
-      id = response["sessionId"]
+      id = response["sessionId"] || get_in(response, ["value", "sessionId"])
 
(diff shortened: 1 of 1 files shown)


michallepicki:
There are other incompatibilities in newer drivers, like for example the element `/displayed` endpoint not being present (one should use javascript for that, as [the spec suggests](https://w3c.github.io/webdriver/#element-displayedness) and reference selenium clients do), hopefully I'll make PRs for these too

keathley:
Thanks @michallepicki! Should we be erroring out if we can't get an id from either of these sources? That way we halt progress and help stop any confusion from the end users perspective? Also is this related to #386?

michallepicki:
After giving it a bit more thought, and learning more about the differences between the JSON Wire Protocol and WebDriver W3C spec, I think sessionID is wrapped in a "value" object only for drivers compatible with the WebDriver W3C spec. Since we can't guarantee our current selenium client will work with such drivers, I think we should raise an error instead.

PR #413, opened 2019-02-05
Handle Safari 11 and Edge 17 stale reference error

michallepicki:
Edge 17 responds with: "Stale element..." (uppercase S)
Safari 11 responds with: "An element command failed..."

### DIFF
--- a/lib/wallaby/httpclient.ex
+++ b/lib/wallaby/httpclient.ex
@@ -77,11 +77,16 @@ defmodule Wallaby.HTTPClient do
     end
   end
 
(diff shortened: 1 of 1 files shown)


keathley:
Looks like this is failing due to a credo check. I personally don't find this code to be unreadable or too complex so I'm fine with adding an ignore for that warning so we can move this forward.

michallepicki:
@keathley right, I'll take a look and try to fix it first

michallepicki:
@keathley You were right, it was simpler to just ignore the warning. Tests passed now

keathley:
Thanks for fixing that up :tada:

Issue #410, opened 2019-02-01
Update elixir versions we test against in travis


keathley:
We need to bump our versions of elixir in travis since we're running a few versions behind at this point

PR #414, opened 2019-02-07
Update .travis.yaml to include the 3 latest elixir versions

saulecabrera:
This PR solves #410 

Just followed the previous convention and added the 3 latest versions.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,8 @@
 language: elixir
 elixir:
-  - 1.4.5
-  - 1.5.3
-  - 1.6.4
+  - 1.6.6
+  - 1.7.4
+  - 1.8.1
 otp_release:
   - 20.3
 env:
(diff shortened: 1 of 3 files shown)


saulecabrera:
Also, should I update `.tool-versions` to match any of the versions that we are testing against?

keathley:
Looks like we've got some issues with deprecation warnings in newer versions of elixir. Need to see if we can safely solve those and still support older versions of the language.

saulecabrera:
Right, I just checked the logs for the builds and the warning relies on the usage of `super` for `handle_info/2` and `terminate/2` in the following files: 
- https://github.com/keathley/wallaby/blob/master/lib/wallaby/driver/process_workspace/server.ex
- https://github.com/keathley/wallaby/blob/master/lib/wallaby/phantom/server.ex#L88

This deprecation was introduced in elixir 1.7 https://github.com/elixir-lang/elixir/blob/v1.7/CHANGELOG.md#elixir-7

Since these are GenServer overridables, if we wanted to be explicit we could return the default implementation for each:
- `{:noreply, state}` for `handle_info/2`
- `:ok` for `terminate/2`

I pushed the changes which I tested locally, let me know if they make sense.

keathley:
That all makes sense to me. Thanks for making those changes @saulecabrera!

Issue #419, opened 2019-03-20
Wallaby failing to start on CircleCi with new chromedriver version


tjshipe:
As of yesterday all of my CircleCi builds are failing when trying to start up Wallaby. I think this has something to do with a new version or versioning system for Chromedriver.

```
** (MatchError) no match of right hand side value: {:error, {:wallaby, {:bad_return, {{Wallaby, :start, [:normal, []]}, {:EXIT, {%Protocol.UndefinedError{description: "", protocol: Enumerable, value: nil}, [{Enumerable, :impl_for!, 1, [file: 'lib/enum.ex', line: 1]}, {Enumerable, :slice, 1, [file: 'lib/enum.ex', line: 193]}, {Enum, :backwards_compatible_slice, 1, [file: 'lib/enum.ex', line: 2871]}, {Enum, :slice_any, 3, [file: 'lib/enum.ex', line: 3075]}, {Enum, :at, 3, [file: 'lib/enum.ex', line: 381]}, {Wallaby.Experimental.Chrome, :validate, 0, [file: 'lib/wallaby/experimental/chrome.ex', line: 36]}, {Wallaby, :start, 2, [file: 'lib/wallaby.ex', line: 27]}, {:application_master, :start_it_old, 4, [file: 'application_master.erl', line: 277]}]}}}}}}
    test/test_helper.exs:1: (file)
    (elixir) lib/code.ex:767: Code.require_file/2
    (elixir) lib/enum.ex:769: Enum."-each/2-lists^foreach/1-0-"/2
    (elixir) lib/enum.ex:769: Enum.each/2
Exited with code 1
```

mhanberg:
What docker container are you using for your Circle CI builds? 

Mine seem to be fine. I'm using `circleci/elixir:1.8-node-browsers`

tjshipe:
Hm, strange. I'm using that same container. It looks like there were a couple issues on circleci-images that could be related...

https://github.com/circleci/circleci-images/issues/355

tjshipe:
I just upgrade my chromedriver locally `chromedriver 2.46 -> 73.0.3683.68` and I'm now seeing the error locally.

```
[...]

PR #420, opened 2019-03-20
Support new chromedriver versioning

tjshipe:
This is to address the issue raised in https://github.com/keathley/wallaby/issues/419

Chromedriver has updated their versioning system to match Chrome's version: http://chromedriver.chromium.org/downloads

Chromedriver 2.xx -> 73.x.xxxx.xx


tjshipe:
doh! Thanks, I'm not sure how I missed that

michallepicki:
@tjshipe No worries, easy one to overlook. Thanks for the PR anyways!

PR #421, opened 2019-03-27
Interacting with opened windows

### LINKED ISSUE #360: Interacting with opened windows/popups
Hi!

I don't really know how to tackle an OAuth authentication step that happens on a popup. With Capybara, for example, you can access the driver and from there its windows. I haven't found a way to do that with Wallaby.

Apart from that, this is an awesome library, thanks!

michallepicki:
Closes #360

TODO:
- [x] Expose through Selenium, Chrome and Browser modules
- [x] Check no window_handle vs "current" window_handle (spec says it's "current"), possibly change get_window_size and set_window_size to use it as well
- ~~Revisit handling errors, reading the value from the response and return types (`:ok` is just the happy path)~~ that's fine for now I think
- [x] Test in Selenium: Safari 11, Edge 17, Firefox, Chrome, Opera, Appium (older version compatible with this protocol)
- [x] Test in Chromedriver with a few Chrome and Chromium versions
- [x] Add documentation

For the unaware: I'm following the [Selenium WebDriver Wire Protocol](https://github.com/SeleniumHQ/selenium/wiki/JsonWireProtocol). The [W3C WebDriver Protocol](https://w3c.github.io/webdriver/#endpoints) has too many changes to try to support both of them at the same time. We will need to have a separate WebdriverClient module for it for people that want to test in Safari 12 or newer, Edge 18 or newer, new Appium versions etc.

### DIFF
--- a/lib/wallaby/experimental/selenium/webdriver_client.ex
+++ b/lib/wallaby/experimental/selenium/webdriver_client.ex
@@ -428,7 +428,7 @@ defmodule Wallaby.Experimental.Selenium.WebdriverClient do
   end
 
   @doc """
-  Closes the current window (or tab)
+  Closes the currently focused window (or tab)
   """
   @spec close_window(Session.t) :: {:ok, map}
   def close_window(session) do
--- a/lib/wallaby/experimental/selenium/webdriver_client.ex
+++ b/lib/wallaby/experimental/selenium/webdriver_client.ex
@@ -366,6 +366,16 @@ defmodule Wallaby.Experimental.Selenium.WebdriverClient do
       do: value
   end
 
+  @doc """
+  Changes focus to specified window
+  """
+  @spec focus_window(Session.t(), String.t()) :: :ok
+  def focus_window(session, window_handle) do
+    with {:ok, _resp} <-
+           request(:post, "#{session.url}/window", %{handle: window_handle, name: window_handle}),
+         do: :ok
+  end
+
   @spec cast_as_element(Session.t | Element.t, map) :: Element.t
   defp cast_as_element(parent, %{"ELEMENT" => id}) do
     cast_as_element(parent, %{@web_element_identifier => id})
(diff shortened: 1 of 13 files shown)


michallepicki:
I think getting and setting window size or position using a handle is not supported in many WebDrivers. Since the user can always change focus to that window first before executing these commands, I'm going to simplify this PR and remove the versions that take a handle. I'll keep the change from "no handle" to "current" to designate the target window since it seems to be working fine and it's more spec compliant with the old protocol.

michallepicki:
@mhanberg @keathley I think this is good enough!

Travis is having some issues with `apt-get` so I'll keep an eye on the build and restart it later.

I didn't add an integration test for `maximize_window` because this command requires a window manager to be running.

A few existing commands got slightly changed to be more in-line with the Selenium WebDriver spec (w3c endpoints were being used), versions that take a window handle got dropped for simplicity, and return values got adjusted so that we always return `:ok` from the WebDriverClient module when the command succeeded.

lib/wallaby/experimental/selenium/webdriver_client.ex
```diff
@@ -350,24 +372,69 @@ defmodule Wallaby.Experimental.Selenium.WebdriverClient do
...
           {:ok, value} <- Map.fetch(resp, "value"),
-      do: value
+      do: {:ok, value}
+  end
+
+  @doc """
+  Changes focus to another window (or tab)
+
+  The window to change focus to may be specified by its server assigned
+  window handle, or by the value of its name attribute.
```
mhanberg: "You may specify the the window by its server-assigned window handle or by the value of its name attribute."

WDYT?

lib/wallaby/experimental/selenium/webdriver_client.ex
```diff
@@ -271,35 +275,52 @@ defmodule Wallaby.Experimental.Selenium.WebdriverClient do
...
+  def get_window_position(session) do
+    with {:ok, resp} <- request(:get, "#{session.url}/window/current/position"),
           {:ok, value} <- Map.fetch(resp, "value"),
       do: {:ok, value}
   end
-  def get_window_size(session, window_handle) do
-    with {:ok, resp} <- request(:get, "#{session.url}/window/#{window_handle}/size"),
+
+  @doc """
+  Maximizes the window if not already maximized.
```
mhanberg:
```suggestion
  Maximizes the window if not already maximized
```

PR #428, opened 2019-05-08
Add support for handling JS confirm, alert and prompt in Selenium

michallepicki:
JS window.alert, window.confirm and window.prompt were handled by Chromedriver and PhantomJS, but for Selenium functions for accept and dismiss them were not implemented.

This PR adds code that allows to use the following functions:

    accept_alert/2,
    accept_confirm/2,
    accept_prompt/2,
    accept_prompt/3,
    dismiss_confirm/2
    dismiss_prompt/2,
    dismiss_alert/2.

To make it work with Firefox as well as Chrome, the `dismiss_prompt/2` endpoint was switched to the one specified by Selenium WebDriver JSON Wire Protocol (instead of the endpoint specified in the W3C WebDriver specification), which is the same as dismiss_confirm/2. This change also affects the Chromedriver backend (since it uses the same WebDriver module as Selenium) but should not cause any regressions.

Since all dialog tests should be now passing on all platforms, `integration_test/cases/browser/dialog_test.exs` was moved to the set of default tests for all drivers.

edit: force-pushed to rebase on current upstream master

### DIFF
--- a/integration_test/chrome/all_test.exs
+++ b/integration_test/chrome/all_test.exs
@@ -1,7 +1,6 @@
 Code.require_file "../tests.exs", __DIR__
 
 # Additional test cases supported by phantom
-Code.require_file "../cases/browser/dialog_test.exs", __DIR__
 Code.require_file "../cases/browser/file_test.exs", __DIR__
 Code.require_file "../cases/browser/js_errors_test.exs", __DIR__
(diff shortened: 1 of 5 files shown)


michallepicki:
Conflict resolved

michallepicki:
Can I get a thumbs up again please? Couldn't merge earlier because of a conflict

keathley:
@michallepicki Done :)

PR #435, opened 2019-07-15
Hover

### LINKED ISSUE #434: create mouse hover action
Thank you for all your work in creating this great library 

Is it possible to add support to move the mouse over elements

Thanks

mhanberg:
This brings support for hovering over elements for Chrome and Selenium.

I attempted to add support for Phantom, but the effort seems better spent on deprecating Phantom instead of adding features to it.

But I admit that it seems misleading considering the current language in the README, which deems Phantom having first class support and selenium/chrome being experimental.

cc/ @keathley @michallepicki 

Resolves #434

### DIFF
--- a/.tool-versions
+++ b/.tool-versions
@@ -1,2 +1,2 @@
-elixir 1.7.4
+elixir 1.7.4-otp-20
 erlang 20.0
--- a/integration_test/cases/browser/hover_test.exs
+++ b/integration_test/cases/browser/hover_test.exs
@@ -8,11 +8,11 @@ defmodule Wallaby.Integration.Browser.HoverTest do
   describe "hover/2" do
     test "sends keys to the specified element", %{page: page} do
       refute page
-      |> visible?(Query.text("HI"))
+             |> visible?(Query.text("HI"))
 
       assert page
-      |> hover(Query.css(".group"))
-      |> visible?(Query.text("HI"))
+             |> hover(Query.css(".group"))
+             |> visible?(Query.text("HI"))
     end
   end
 end
--- a/integration_test/cases/browser/hover_test.exs
+++ b/integration_test/cases/browser/hover_test.exs
@@ -1,0 +1,18 @@
+defmodule Wallaby.Integration.Browser.HoverTest do
+  use Wallaby.Integration.SessionCase, async: true
+
+  setup %{session: session} do
+    {:ok, page: visit(session, "hover.html")}
+  end
+
+  describe "hover/2" do
+    test "sends keys to the specified element", %{page: page} do
+      refute page
+      |> visible?(Query.text("HI"))
+
+      assert page
+      |> hover(Query.css(".group"))
+      |> visible?(Query.text("HI"))
+    end
+  end
+end
--- a/integration_test/cases/element/hover_test.exs
+++ b/integration_test/cases/element/hover_test.exs
@@ -1,0 +1,22 @@
+defmodule Wallaby.Integration.Element.HoverTest do
+  use Wallaby.Integration.SessionCase, async: true
+
+  setup %{session: session} do
+    {:ok, page: visit(session, "hover.html")}
+  end
+
+  describe "hover/2" do
+    test "sends keys to the specified element", %{page: page} do
+      page
+      |> find(Query.text("HI", visible: false), fn el ->
+        refute Element.visible?(el)
+      end)
+      |> find(Query.css(".group"), fn el -> 
+        Element.hover(el)
+      end)
+      |> find(Query.text("HI"), fn el ->
+        assert Element.visible?(el)
+      end)
+    end
+  end
+end
--- a/integration_test/cases/element/hover_test.exs
+++ b/integration_test/cases/element/hover_test.exs
@@ -11,7 +11,7 @@ defmodule Wallaby.Integration.Element.HoverTest do
       |> find(Query.text("HI", visible: false), fn el ->
         refute Element.visible?(el)
       end)
-      |> find(Query.css(".group"), fn el -> 
+      |> find(Query.css(".group"), fn el ->
         Element.hover(el)
       end)
       |> find(Query.text("HI"), fn el ->
--- a/lib/wallaby/driver/log_store.ex
+++ b/lib/wallaby/driver/log_store.ex
@@ -8,6 +8,10 @@ defmodule Wallaby.Driver.LogStore do
   @doc """
   Appends logs to a session
   """
+  def append_logs(session, logs) when is_binary(session) and not is_list(logs) do
+    append_logs(session, List.wrap(logs))
+  end
+
   def append_logs(session, logs) when is_binary(session) do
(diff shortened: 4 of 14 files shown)


.tool-versions
```diff
@@ -1,2 +1,2 @@
-elixir 1.7.4
+elixir 1.7.4-otp-20
```
michallepicki: This is to start using the version used on Travis, I presume? Looks like we're using elixir precompiled with otp 20 there indeed:
```
$ elixir --version
Erlang/OTP 20 [erts-9.3] [source] [64-bit] [smp:2:2] [ds:2:2:10] [async-threads:10] [hipe] [kernel-poll:false]
Elixir 1.7.4 (compiled with Erlang/OTP 20)
```
If that's the goal we can also set erlang to 20.3 in the following line

integration_test/cases/browser/hover_test.exs
```diff
@@ -0,0 +1,18 @@
+defmodule Wallaby.Integration.Browser.HoverTest do
+  use Wallaby.Integration.SessionCase, async: true
+
+  setup %{session: session} do
+    {:ok, page: visit(session, "hover.html")}
+  end
+
+  describe "hover/2" do
+    test "sends keys to the specified element", %{page: page} do
```
michallepicki: Should we change the test name to something more fitting here?

integration_test/cases/element/hover_test.exs
```diff
@@ -0,0 +1,22 @@
+defmodule Wallaby.Integration.Element.HoverTest do
+  use Wallaby.Integration.SessionCase, async: true
+
+  setup %{session: session} do
+    {:ok, page: visit(session, "hover.html")}
+  end
+
+  describe "hover/2" do
+    test "sends keys to the specified element", %{page: page} do
```
michallepicki: Should we change the test name to something more fitting here?

lib/wallaby/driver/log_store.ex
```diff
@@ -8,6 +8,10 @@ defmodule Wallaby.Driver.LogStore do
   @doc """
   Appends logs to a session
   """
+  def append_logs(session, logs) when is_binary(session) and not is_list(logs) do
```
michallepicki: Not sure why are we making this change? Seems unrelated to hover

lib/wallaby/phantom.ex
```diff
@@ -116,59 +116,59 @@ defmodule Wallaby.Phantom do
   end
 
   @doc false
-  defdelegate accept_alert(session, open_dialog_fn),              to: Driver
```
michallepicki: Not sure why are we making these changes? Seems unrelated to hover

test/wallaby/phantom/log_store_test.exs
```diff
@@ -9,12 +9,21 @@ defmodule Wallaby.Driver.LogStoreTest do
 
   @session "123abc"
 
+  setup do
```
michallepicki: Not sure why are we making this change? Seems unrelated to hover

.tool-versions
mhanberg: It's not for Travis, it's for asdf. I think elixir 1.7.4 comes compiled for OTP 21 by default, which causes it to crash immediately.

So you have to explicitly use as compiled with otp-20 for it to work at all.

lib/wallaby/driver/log_store.ex
mhanberg: It is a bug that I discovered while TDDing this feature. It errors when attempting to remove duplicate logs because it calls `--` with a list and a map instead of two lists.

I have also seen this bug in travis logs.

lib/wallaby/phantom.ex
mhanberg: I thought it was hard to read and was better not formatted that way ¯\_(ツ)_/¯.

test/wallaby/phantom/log_store_test.exs
mhanberg: See [previous comment](https://github.com/keathley/wallaby/pull/435#discussion_r304444908)

integration_test/cases/browser/hover_test.exs
mhanberg: Whoops, yeah I had copied the test structure from the "send_keys" tests and forgot to rename. 👍

mhanberg:
@michallepicki Thanks for the review 👌.

.tool-versions
michallepicki: I just checked, asdf elixir 1.7.4 comes precompiled using OTP 19:
```
elixir --version
Erlang/OTP 21 [erts-10.3.1] [source] [64-bit] [smp:6:6] [ds:6:6:10] [async-threads:1] [hipe]

Elixir 1.7.4 (compiled with Erlang/OTP 19)
```
This shouldn't be a problem, I didn't see any crashes myself? But we could make this change I think, it will make the version the same as on travis

lib/wallaby/phantom.ex
michallepicki: Oh it's just formatting? That's alright

lib/wallaby/driver/log_store.ex
michallepicki: Can you send a link to the travis failure on master branch that this change fixes?

mhanberg: https://travis-ci.org/keathley/wallaby/jobs/559192090

```
01:05:02.545 [error] GenServer Wallaby.Driver.LogStore terminating
** (ArgumentError) argument error
    :erlang.--(%{"error" => "invalid session id", "message" => "invalid session id", "stacktrace" => "#0 0x55ee2df547a9 <unknown>\n"}, [%{"level" => "SEVERE", "message" => "http://localhost:41715/images/button-image.png - Failed to load resource: the server responded with a status of 404 (Object Not Found)", "source" => "network", "timestamp" => 1563239096742}, %{"level" => "SEVERE", "message" => "http://localhost:41715/images/button-image.png - Failed to load resource: the server responded with a status of 404 (Object Not Found)", "source" => "network", "timestamp" => 1563239096781}])
```

michallepicki: I looked through it and the patch makes sense, although I think this could be a separate PR for easier review ;)

test/wallaby/phantom/log_store_test.exs
michallepicki: Looks good, but can we keep the call the same? I think LogStore being an Agent is an implementation detail and we can keep LogStore.start_link here

mhanberg: The call to `LogStore.start_link` was actually silently (silent because it wasn't matching on anything) failing because `LogStore` is actually started under a supervisor when the app starts up. 

The state needs to be cleared somehow between tests or the implementation of LogStore needs to change to take a name so it can be started independently while under test.

In general I agree with you though. I will figure something out.

cysiegel:
@marcdel @michallepicki Thanks for doing this so quickly!!!!  
Wallaby is great!

PR #439, opened 2019-07-20
Run selenium tests in headless mode in Firefox

michallepicki:
Chrome tests are being run in headless mode, so I think we can do that for selenium as well. This is less annoying when using selenium-standalone jar and locally installed firefox, because newly opened windows were stealing system focus.


michallepicki:
This doesn't change the default way user tests are being run, this change affects only our internal wallaby integration tests

PR #440, opened 2019-07-20
Fix Wallaby.start_session_opts type spec - passing :driver doesn't do anyting

michallepicki:
See https://github.com/keathley/wallaby/blob/master/lib/wallaby.ex#L80
While I agree that the `:driver` option could be useful, currently it's not implemented.


michallepicki:
There's a conflict, I'll rebase

michallepicki:
@mhanberg Nice find, done!

PR #441, opened 2019-07-22
Enable running hover integration tests added in #435

michallepicki:
@mhanberg Tests weren't added to  Chrome and Selenium lists for tests to run


mhanberg:
🤦‍♂ good catch.

PR #443, opened 2019-07-23
Update and add new elixir and erlang versions used for testing on travis

michallepicki:
This would be 5 * 3 * 3 = 45 jobs to run so I'm not sure if it's a good idea. Maybe we could reduce the amount of jobs to run for one version combo?

edit: Dropped selenium 2, dropped elixir 1.7 and otp 20, also excluded elixir 1.8+otp22 and elixir 1.9+otp21 to simplify the matrix. Now it's 2x4, looks good but I need to fix the dialyzer warning

### DIFF
--- a/.tool-versions
+++ b/.tool-versions
@@ -1,2 +1,2 @@
-elixir 1.7.4-otp-20
-erlang 20.0
+elixir 1.8.2-otp-21
+erlang 21.3
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,12 @@
 language: elixir
 elixir:
   - 1.7.4
-  - 1.8.1
+  - 1.8.2
+  - 1.9.1
 otp_release:
   - 20.3
+  - 21.3
+  - 22.0
 env:
   - ""
   - WALLABY_DRIVER=phantom
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,6 @@ matrix:
 sudo: required
 cache:
   directories:
-    - _build
     - deps
 
 dist: trusty
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,18 +1,21 @@
 language: elixir
 elixir:
-  - 1.7.4
   - 1.8.2
   - 1.9.1
 otp_release:
-  - 20.3
   - 21.3
   - 22.0
 env:
   - ""
   - WALLABY_DRIVER=phantom
   - WALLABY_DRIVER=chrome
-  - WALLABY_DRIVER=selenium WALLABY_SELENIUM_VERSION=3
-  - WALLABY_DRIVER=selenium WALLABY_SELENIUM_VERSION=2
+  - WALLABY_DRIVER=selenium
+matrix:
+  exclude:
+  - elixir: 1.8.2
+    otp_release: 22.0
+  - elixir: 1.9.1
+    otp_release: 21.3
 # Run in bigger container so we don't run out of memory generating the plt
 sudo: required
 cache:
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,6 +20,7 @@ matrix:
 sudo: required
 cache:
   directories:
+    - _build
     - deps
 
 dist: trusty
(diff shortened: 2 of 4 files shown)


keathley:
I can't recall a time when the matrix has found an issue between elixir and erlang versions. I think it's safe to pin a major version of elixir to a major erlang release.

mhanberg:
I personally think it's okay to do

mhanberg:
I personally think it's okay to do:

- latest two elixir versions
- whatever is the supported OTP version (21 currently i believe)

and also drop Selenium 2 (now) and Phantom (when the time comes).

michallepicki:
I don't have a clue why we're seeing [this](https://travis-ci.org/keathley/wallaby/jobs/563005494#L888) dialyzer warning: 
```
lib/wallaby/experimental/chrome/logger.ex:35: The pattern 
          {'error', _} can never match the type 
          {'ok',
           'false' | 'nil' | 'true' |
           binary() |
           [any()] |
           number() |
           map()}
```
It points to[here](https://github.com/keathley/wallaby/blob/master/lib/wallaby/experimental/chrome/logger.ex#L35):
```
    case Poison.decode(unescaped) do
      {:ok, data} -> "\n#{Poison.encode!(data, pretty: true)}"
      {:error, _} -> unescaped
    end
```
`Poison.decode` has a spec of
```
  @spec decode(iodata, Keyword.t) :: {:ok, Parser.t} | {:error, :invalid}
    | {:error, {:invalid, String.t}}
```
where `Parser.t` is `nil | true | false | list | float | integer | String.t | map`. So for some reason dialyzer in newer Elixir/Erlang versions misreads the type for `Poison.decode` omitting half of what's specified there?

michallepicki:
Anyway, would it be a good idea to move from Poison to Jason, or give users the choice?

keathley:
I think its safe to move to jason at this point.

keathley:
I think we're still locked to a really old version of poison as well so even if we stick with poison its probably worth upgrading.

michallepicki:
I checked that and we're on poison 3.1 which is the latest that most people use ([this commit](https://github.com/devinus/poison/commit/a4208a6252f4e58fbcc8d9fd2f4f64c99e974cc8#comments) fixes our dialyzer issue but also makes poison incompatible with ecto) so if we stick to poison maybe we should just ignore the warning? I don't see anything wrong with our code. Or maybe there's a way to rewrite this code to persuade dialyzer that it's fine?

mhanberg:
Personally, I think we should up move to Jason.

michallepicki:
We can make a separate issue to track that and ignore the warning for now so we can move forward with this PR

mhanberg:
Sounds good 👍

.tool-versions
```diff
@@ -1,2 +1,2 @@
-elixir 1.7.4-otp-20
-erlang 20.0
+elixir 1.8.2-otp-21
```
mhanberg: I ended up figuring out my problem regarding the `-otp-20` need. I think I installed elixir 1.7.4 before asdf started compiling it with opt 19 be default. Just needed to reinstall and then it was fixed.

We can remove the otp suffix if you want, but it doesn't really hurt anything by being there.

.travis.yml
```diff
@@ -1,15 +1,21 @@
...
   - WALLABY_DRIVER=chrome
-  - WALLABY_DRIVER=selenium WALLABY_SELENIUM_VERSION=3
-  - WALLABY_DRIVER=selenium WALLABY_SELENIUM_VERSION=2
+  - WALLABY_DRIVER=selenium
+matrix:
+  exclude:
+  - elixir: 1.8.2
+    otp_release: 22.0
+  - elixir: 1.9.1
+    otp_release: 21.3
```
mhanberg: I think I'm okay with leaving these as part of the matrix.

.tool-versions
michallepicki: I kept the suffix so that we run the same version in dev as on Travis: [link](https://travis-ci.org/keathley/wallaby/jobs/563091366#L456)
Looks like asdf compiles 1.8.2 with otp 20 by default:
```
$ asdf install
elixir 1.8.2 is already installed
erlang 21.3 is already installed
$ elixir -v
Erlang/OTP 21 [erts-10.3] [source] [64-bit] [smp:6:6] [ds:6:6:10] [async-threads:1] [hipe]

Elixir 1.8.2 (compiled with Erlang/OTP 20)
```

.travis.yml
michallepicki: I excluded these as suggested by @keathley [here](https://github.com/keathley/wallaby/pull/443#issuecomment-514198815), and I prefer less waiting :)

PR #374, opened 2018-06-08
Execute script async

daveed-al:
Adding async functionality to executing javascript.

PR #445, opened 2019-07-26
Execute async script

mhanberg:
This is a replacement PR for #374.

It simply merges in master to get it up to date.

It was previously failing the build due to Selenium 2.x, but we recently dropped Selenium 2.x from our CI builds so now it is passing.

There was some worry in #374 about still wanting to support Selenium 2 due to supporting older Firefox versions, but I'm not so worried about that. The [3.0.0](https://github.com/SeleniumHQ/selenium/releases/tag/selenium-3.0.0) release seems to imply that Selenium 3 still works works older versions, but doesn't use GeckoDriver in those cases.

cc/ @michallepicki @keathley


mhanberg:
@michallepicki do you mind reviewing this?

michallepicki:
I'll try to do review it over the weekend, I actually have a use-case in my tests and want to try how using `execute_script_async` would improve my code (currently I'm callling `execute_script` checking value repeatedly)

lib/wallaby/experimental/selenium/webdriver_client.ex
```diff
@@ -336,6 +336,17 @@ defmodule Wallaby.Experimental.Selenium.WebdriverClient do
       do: {:ok, value}
   end
 
+  @doc """
+  Executes javascript asynchronously, taking as arguments the script to execute,
```
michallepicki: I looked into it and I don't think it executes the script `asynchronously`, when executing the script the WebDriver client will block until the callback is called. Maybe we could reword this?

lib/wallaby/browser.ex
```diff
@@ -410,6 +410,34 @@ defmodule Wallaby.Browser do
     parent
   end
 
+  @doc """
+  Executes javascript asynchronously, taking as arguments the script to execute,
```
michallepicki: I looked into it and I don't think it executes the script asynchronously, when executing the script the WebDriver client will block until the callback is called. Maybe we could reword this?

lib/wallaby/experimental/chrome.ex
```diff
@@ -210,6 +210,21 @@ defmodule Wallaby.Experimental.Chrome do
     end
   end
 
+  @doc false
+  def execute_script_async(session_or_element, script, args \\ [], opts \\ []) do
```
michallepicki: I don't really get how logging stuff works and don't use it currently, but looks the same as `execute_script` so :man_shrugging:

PR #446, opened 2019-07-26
Replace Poison with Jason, closes #444

### LINKED ISSUE #444: Possibly move from Poison to Jason
When upgrading to Elixir 1.8.2-otp-21 and OTP 21.3 as minumum versions used for Wallaby development and testing in #443, a new dialyzer warning surfaced:
```
lib/wallaby/experimental/chrome/logger.ex:35: The pattern 
          {'error', _} can never match the type 
          {'ok',
           'false' | 'nil' | 'true' |
           binary() |
           [any()] |
           number() |
           map()}
```
It points to Logger:
```
    case Poison.decode(unescaped) do
      {:ok, data} -> "\n#{Poison.encode!(data, pretty: true)}"
      {:error, _} -> unescaped
    end
```
`Poison.decode` has a spec of
```
  @spec decode(iodata, Keyword.t) :: {:ok, Parser.t} | {:error, :invalid}
    | {:error, {:invalid, String.t}}
```
where `Parser.t` is `nil | true | false | list | float | integer | String.t | map`. So for some reason dialyzer in newer Elixir/Erlang versions misreads the type for `Poison.decode` omitting half of what's specified there? It's either a bug in dialyzer or Poison. Anyway I don't see anything wrong with our code.

I don't think we should upgrade Poison because 3.1 is the latest that most people use. [This commit](https://github.com/devinus/poison/commit/a4208a6252f4e58fbcc8d9fd2f4f64c99e974cc8#comments) fixes our dialyzer issue but also makes poison incompatible with ecto.

We can move to Jason from Poison if we don't find a way to fix this warning any other way.

michallepicki:


michallepicki:
One of the tests was failing due to a change of key order in JSON encoding, I updated this test

PR #447, opened 2019-07-30
Wallaby.Feature

### LINKED ISSUE #110: Better names for screenshots
I have configured wallaby to take screenshots when something went wrong.

This was fine when only a single test failed, but when there are multiple ones failing I do have problems to assign the correct screenshot to the test.

So it would be nice if either the testsname or some info from `conn` (`"#{method}-#{path}"`) could be part of the name of the screenshot.

### LINKED ISSUE #142: Feature Case for exunit
We should include a `feature` exunit test type. Having a custom type will allow us to capture more information and use that information to create better screenshot names and capture screenshots on assertion failures.

### LINKED ISSUE #471: screenshot_on_failure not being called for `assert_text/2`
## Issue

Hi! 

I'm trying to get the `screenshot_on_failure` feature working, but I can't get it to work with `assert_text/2`. This happens because the `/2` version of the function does not call the `take_screenshot` one, but the `/3` version does (via `find`):

https://github.com/keathley/wallaby/blob/e15c8b2cf8d4b444a4dea9fd1f33eca161348d15/lib/wallaby/browser.ex#L799-L806

I'll try to submit a PR fixing it!

mhanberg:
It would be useful to have a built in ExUnit case that works for most Wallaby users.

This is a first draft in an attempt to gather feedback to ensure that I was going down the right path.

cc/ @keathley @michallepicki 

- Creates `feature` test macro to replace `test`.
    - Takes screenshot on any feature failure, not just in calls to `find` and `assert_has`
    - Removes screenshotting from other call sites, as the above makes them obsolete.
    - Screenshots are named with the current time, feature name, and the session number (if a feature uses 2 sessions, the respective sessions numbers will be `(1)` `(2)`).
- Checks out all Ecto repos and puts them in shared mode if the test is `async`.
    - This requires the user to configure the `:otp_app` with wallaby (`config :wallaby, :otp_app, :my_app`)
- Registers a `sessions` test attribute that can be used to configure the number of sessions that a feature will start.
    - Defaults to 1 if not set
    - Sessions are available in the `sessions` key of the context passed to the feature (always as a list).

- ~Also uses a `:capabilities` configuration option. Since starting the session is hidden to the user when using `FeatureCase`, there should be a way to expose those options as configuration.~
    - ~Selenium won't work without providing at least a `:browserName` capability, so this is added during the test. There might be cause to pick a browser for Selenium to use by default.~

Resolves #142 
Resolves #110
Resolves #471 

## Questions

- ~How do we address the `phoenix_ecto` and `ecto_sql` dependencies? Compilations will fail if they are run with `--warnings-as-errors`~ A: Conditionally define the modules if they are not loaded.

## Example

```elixir
defmodule MyApp.SomeFeatureTest do
  use Wallaby.FeatureCase, async: true


  @sessions 2
  feature "multi session", %{sessions: [session_1, session_2]} do
    session_1
    |> visit("/page_1.html")
    |> find(Query.css("body > h1"), fn el ->
      assert Element.text(el) == "Page 1"
    end)

    session_2
    |> visit("/page_2.html")
    |> find(Query.css("body > h1"), fn el ->
      assert Element.text(el) == "Page 2"
    end)
  end

  feature "single session", %{sessions: [only_session]} do
    only_session
    |> visit("/page_1.html")
    |> find(Query.css("body > h1"), fn el ->
      assert Element.text(el) == "Page 1"
    end)
  end
end
```

```elixir
# test.exs
config :wallaby,
  otp_app: :my_app,
  screenshot_on_failure: true,
  capabilities: %{
    browserName: "firefox",
    "moz:firefoxOptions": %{
      args: ["-headless"]
    }
  }
```


lib/wallaby/feature_case.ex
```diff
@@ -0,0 +1,90 @@
...
+
+    unless async, do: Ecto.Adapters.SQL.Sandbox.mode(repo, {:shared, self()})
+
+    repo
+  end
+
+  def metadata_for_ecto_repos([]), do: Map.new()
+  def metadata_for_ecto_repos(repos), do: Phoenix.Ecto.SQL.Sandbox.metadata_for(repos, self())
+
+  defmacro feature(message, context \\ quote(do: _), contents) do
```
mhanberg: There is the potential to make this compatible with the following syntax:

```elixir
feature "some feature test" do
  raise "oopsie"
rescue
  # stuff
end
```

the `property` macro from stream_data has the functionality, I'm not sure how useful it would be to have in wallaby.

lib/wallaby/feature_case.ex
```diff
@@ -0,0 +1,90 @@
...
+      ExUnit.Case.register_attribute(__MODULE__, :sessions)
+
+      use Wallaby.DSL
+      import Wallaby.FeatureCase
+    end
+  end
+
+  setup context do
+    sessions_count = get_in(context, [:registered, :sessions]) || 1
+    capabilities = Application.get_env(:wallaby, :capabilities)
```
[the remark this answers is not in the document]
michallepicki: Or don't pass capabilities in start_session at all and let the driver read it from context if it's present

lib/wallaby/feature_case.ex
```diff
@@ -0,0 +1,90 @@
...
+    metadata =
+      @otp_app
+      |> ecto_repos()
+      |> Enum.map(&checkout_ecto_repos(&1, context[:async]))
+      |> metadata_for_ecto_repos()
+
+    sessions =
+      1..sessions_count
+      |> Enum.map(fn _ ->
+        context = String.replace(to_string(context.test), " ", "_")
```
michallepicki: It took me a while to see that `session.context` is just modified test message and not the whole context. Is there a different name we could use?

Also if this will be used in screenshot file name, I think there may be some illegal characters in test message, we could try to handle that

michallepicki:
I think this makes `screenshot_on_failure` more useful.

I have a similar macro in my feature case and I'm using some additional `try-rescue` clauses, e.g. for setting test result to "failed" on Browserstack. It would be nice to provide some way for the user to do that.

I must say though that there are some caveats with this approach. This code won't run (e.g. a screenshot won't be taken) if the test is timing out or some code in user's `setup` block crashes.

Also I think Wallaby sometimes doesn't always properly close the session when a test times out - I had this happen but it's hard to reproduce.

I looked through other options:
-  [on_exit](https://hexdocs.pm/ex_unit/ExUnit.Callbacks.html#on_exit/2) wouldn't work because callback function doesn't take any arguments
- I'm thinking about writing a [ExUnit.Formatter](https://hexdocs.pm/ex_unit/ExUnit.Formatter.html) that would handle screenshots, closing the session or other stuff (like Browserstack status). This may work

lib/wallaby/feature_case.ex
mhanberg: I think renaming is an okay option. I just followed the naming theme from #156 without thinking about it much.

I will look into the illegal characters.

lib/wallaby/feature_case.ex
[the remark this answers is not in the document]
mhanberg: I think these are fine options. I'll explore this a bit more.

mhanberg:
> I must say though that there are some caveats with this approach. This code won't run (e.g. a screenshot won't be taken) if the test is timing out or some code in user's `setup` block crashes.

I think this is a caveat no matter the solution. If the test doesn't start or the web driver crashes/times out, taking a screenshot is sort of not applicable.

> Also I think Wallaby sometimes doesn't always properly close the session when a test times out - I had this happen but it's hard to reproduce.

Is this related to the FeatureCase or a side note? This sounds like a general problem.

> * [on_exit](https://hexdocs.pm/ex_unit/ExUnit.Callbacks.html#on_exit/2) wouldn't work because callback function doesn't take any arguments

`on_exit` isn't viable for taking screenshots, but could be used to ensure that sessions are closed after a test.

> * I'm thinking about writing a [ExUnit.Formatter](https://hexdocs.pm/ex_unit/ExUnit.Formatter.html) that would handle screenshots, closing the session or other stuff (like Browserstack status). This may work

I am not familiar with ExUnit.Formatter, but I fear it would have the same limitations that `on_exit` would have. `on_exit` runs in a separate process and runs after the test process has already ended. You will have already lost the reference to the session or the session will have already been closed.

mhanberg:
Pausing work on this pending #460

mhanberg:
@michallepicki @aaronrenner I think this is ready for another review.

## Updates

### Capabilities

- Since #467 has been merged, there are default capabilities for all of the drivers and has made this implementation simpler.
- You can now pass a list of `Wallaby.start_session` options to the `session` attribute to configure the capabilities for a specific feature. I chose to have it take the `start_session` options rather than just the capabilities because this allows to pass the `create_session_fn` which makes testing possible. I honestly don't imagine this being used too often, but I think it is a reasonable escape hatch.

### Automatic Screenshots

I am not sure how to test this, given it would require a failing test to determine if it works. I have run them manually however and have determined that it works.

## Elixir/Erlang versions

This PR raises the minimum Elixir version to v1.7 and the Erlang version to v21. This is to be able to take advantage of the `__STACKTRACE__ special form, which was introduced in those versions. 

This also decreases the CI time 😏.

## Ignored Dialyzer warnings and Credo rule change

There have been several warnings added to the ignore file. I figured they had something to do with Dialyzer not understanding Elixir macros so I decided it wasn't work the effort to figure out.

I turned off the `AliasUsage` Credo rule because 1) I don't like that rule too much and 2) it was giving some false positives inside the `feature` macro.

aaronrenner:
I like how this PR is trying to take care of some common problems users run into like taking screenshots on test failure and checking out ecto sandboxes. Would a FeatureCase be usable if I'm not using ecto? I see it adds a dependency on `ecto_sql` and `phoenix_ecto`, which may not be needed if someone is using wallaby for smoke testing or scraping non-phoenix websites.

mhanberg:
@aaronrenner it should be usable, but I am not totally sure how `:optional` dependencies work. The tests run without them as dependencies, but there are compiler warnings about the modules not being defined, which causes CI to break due to treating warnings as errors.

It is definitely not okay to install those dependencies into the users project if they aren't using them.

This is reminds me that I still need to test this branch in an actual project that uses phoenix and ecto to make sure the ecto repo stuff is working.

>I like how this PR is trying to take care of some common problems users run into like taking screenshots on test failure and checking out ecto sandboxes

Yes, my goal with this is to make it easier to start using Wallaby without as much up front setup.

lib/wallaby/feature_case/ecto_sql/sandbox.ex
```diff
@@ -0,0 +1,8 @@
+unless Code.ensure_loaded?(Ecto.Adapters.SQL.Sandbox) do
+  defmodule Ecto.Adapters.SQL.Sandbox do
+    @moduledoc false
+
+    def checkout(_repo), do: nil
+    def mode(_repo, _mode), do: nil
+  end
+end
```
mhanberg: @aaronrenner I think this should alleviate the following concerns. What do you think?

>Would a FeatureCase be usable if I'm not using ecto? I see it adds a dependency on ecto_sql and phoenix_ecto, which may not be needed if someone is using wallaby for smoke testing or scraping non-phoenix websites.

mhanberg:
> @aaronrenner it should be usable, but I am not totally sure how `:optional` dependencies work. The tests run without them as dependencies, but there are compiler warnings about the modules not being defined, which causes CI to break due to treating warnings as errors.
> 
> It is definitely not 
> This is reminds me that I still need to test this branch in an actual project that uses phoenix and ecto to make sure the ecto repo stuff is working.

lib/wallaby/feature_case/ecto_sql/sandbox.ex
aaronrenner: Instead of defining fake replacement modules, could there be a flag in the feature case that disables the Ecto sandbox integration?

mhanberg: What are you imagining?

I'd like to clarify that if the user doesn't configure their `otp_app`, the ecto stuff has no effect on their tests. 

The feature case still makes automatic screenshots more useful (taking them on any test failure and providing a name that is based on the test's name) and making it easier to create a feature out of the box (the user doesn't have to create their own feature case).

lib/wallaby/feature_case.ex
```diff
@@ -0,0 +1,174 @@
...
+
+    context = Macro.escape(context)
+    contents = Macro.escape(contents, unquote: true)
+
+    quote bind_quoted: [context: context, contents: contents, test_name: test_name] do
+      name = ExUnit.Case.register_test(__ENV__, :feature, test_name, [:feature])
+
+      def unquote(name)(unquote(context)), do: unquote(contents)
+    end
+  end
```
aaronrenner: I did some research and this approach seems to make sense to take automatic screenshots. 👍 

Instead of requiring sessions are created with the `@sessions` attribute, we might be able to leverage the `Wallaby.SessionStore` that's responsible for shutting down sessions when the test ends. I noticed that `Wallaby.SessionStore` is already keeping track of the sessions that have been opened by a given process and wonder if we could pull the sessions that need to be screenshotted out of there. 🤔 

If that works, the only change the user would need to make to their code for automatic screenshots, would be to change the `test` macro to `feature`. This would allow them to continue creating sessions with `Wallaby.start_session/1` like they always have. Do you think this might be a possibility?

mhanberg: I just realized that the response that I thought had commented here last night never sent... and I lost the text.

>I noticed that Wallaby.SessionStore is already keeping track of the sessions that have been opened by a given process and wonder if we could pull the sessions that need to be screenshotted out of there.

I think the SessionStore might be a viable location to take screenshots. It depends if we can differentiate between a failed test and a successful test when the SessionStore receives the `:DOWN` message. 

>If that works, the only change the user would need to make to their code for automatic screenshots, would be to change the test macro to feature.

I don't think it would matter which macro they use if we were able to make the above work.

>This would allow them to continue creating sessions with Wallaby.start_session/1 like they always have

I don't think that taking the screenshots with the SessionStore (while still using the `feature` macro) like that would allow for this. I also don't think allowing for this is beneficial, outside of easing the upgrade path. 

If we could take screenshots in the SessionStore, they would gain better screenshots without having to change their tests.

---

Let me know if you have any more questions @aaronrenner so we can get this merged 👍.

aaronrenner: I didn't actually mean that we'd have the SessionStore take the screenshot. SessionStore would be more for tracking open sessions and we'd have another process that would take screenshots for each of the open sessions on test failure. I have some time during the day tomorrow that I could put together a prototype and get your thoughts if you like.

mhanberg: Yeah, sorry I didn't mean that process would literally take the screenshot, but what I'm picturing would have the SessionStore delegating the taking of the screenshot at least.

aaronrenner: @mhanberg Here's a small proof of concept for what I was talking about: https://github.com/keathley/wallaby/compare/master...aaronrenner:ar-screenshot-on-error. If we track can grab the sessions that we're opened by the current process from `Wallaby.SessionStore` the test macro can then take screenshots on failure.

The tests would look like this.
```elixir
defmacro MyApp.MyFeature do
  use ExUnit.Case
  use Wallaby.DSL
  
  import Wallaby.FeatureTest 

  feature "a failing feature that takes two screenshots" do
    Wallaby.start_session()
    Wallaby.start_session()

    # 2 screenshots are automatically taken when this fails
    assert false
  end
end
```

I copied the feature macro from what you had, and just tweaked the session store. It also keeps the migration path straight forward for the users...

  1. `import Wallaby.FeatureTest` (named after `Phoenix.ConnTest`, but I'm not tied to the name.
  2. Change the `test` macro to `feature`

I know this doesn't do everything your current PR does, but I think it's pretty light weight. What do you think about this approach?

mhanberg:
@aaronrenner @michallepicki Disregarding the tests not passing, I think this is ready for another review.

I used a lot of Aaron's feedback from the last time around and incorporated some of the code that he showed me.

This has transformed from `Wallaby.FeatureCase` to `Wallaby.Feature`. It used to be an `ExUnit.CaseTemplate`, where as now it is just a module. There was not really any reason for it to be a CaseTemplate.

One of the main differences now is that the user can either `use` the module or `import` the module.

`use`ing the module will import the `feature` macro as well as automagically check out their ecto repos if they have them and start a session for them.

`import`ing the module will only include the `feature` macro. 

In both cases, screenshots will automatically be taken and named appropriately based on the feature.

I think I can still improve the documentation and the README, but I wanted to see what y'all think of this so far.

### Other

A couple other things included in this are

- change CI to only build the `pull_request` trigger if its from a fork. This way when we open PRs, it won't do a double build.
- Remove the old docs alias preferred env

.github/workflows/ci.yml
```diff
@@ -6,8 +6,10 @@ env:
 
 jobs:
   linting:
-    runs-on: ubuntu-latest
     name: Linting (${{matrix.elixir}}/${{matrix.otp}})
+    runs-on: ubuntu-latest
+
+    if: github.event_name == 'push' || (github.event_name == 'pull_request' && github.event.pull_request.head.repo.fork == true)
```
aaronrenner: Good find on this! Would you mind submitting this as a separate PR so we can use it immediately?

mix.exs
```diff
@@ -60,8 +60,8 @@ defmodule Wallaby.Mixfile do
       {:credo, "~> 0.9", only: [:dev, :test], runtime: false},
       {:bypass, "~> 1.0.0", only: :test},
       {:excoveralls, "~> 0.7", only: :test},
-      {:ex_doc, "~> 0.20", only: :docs},
-      {:inch_ex, "~> 2.0", only: :docs}
+      {:ex_doc, "~> 0.20", only: :dev},
+      {:inch_ex, "~> 2.0", only: :dev}
```
aaronrenner: Just curious... why remove the `:docs` environment? I first saw this pattern with phoenix and thought it made sense because `ex_doc` and `inch_ex` aren't needed during development, only when generating documentation.

https://github.com/phoenixframework/phoenix/blob/3d964c67c7bf4c282bce36fca57bf990691a222a/mix.exs#L77-L79

.github/workflows/ci.yml
mhanberg: Yep, no problem

mix.exs
mhanberg: TBH, I forgot that I had switched the actual dependencies back to `:dev` in this PR, I was thinking I just switched the `preferred_cli_env`. 

I thought this was originally made this way due to an issue with the Elixir version we were using. I don't feel super strongly about it, I can switch it back.

I should also set the `preferred_cli_env` for the `mix inch` tasks so we don't have to worry about setting the env there either.

michallepicki: I think this simplifies things a bit? I wouldn't mind if they get downloaded when I'm working on wallaby.

That said, they still won't get downloaded for wallaby users in dev env, right? This would be less ideal, but I'm not sure how that works, I'd have to test it out.

mhanberg: Does the env actually change which deps are downloaded or just which ones are loaded?

aaronrenner: I was just playing around a little and found it doesn't change which deps are downloaded by `mix deps.get` (unless you use `--only dev`). However when I run `mix deps.compile`, if these documentation libraries are in a different MIX_ENV, it won't compile or load these libraries or their dependencies: 

```
$ mix deps.tree ex_doc
ex_doc ~> 0.20 (Hex package)
├── earmark ~> 1.3 (Hex package)
└── makeup_elixir ~> 0.14 (Hex package)
    └── makeup ~> 1.0 (Hex package)
        └── nimble_parsec ~> 0.5.0 (Hex package)

$ mix deps.tree inch_ex
inch_ex ~> 2.0 (Hex package)
├── bunt ~> 0.2 (Hex package)
└── jason ~> 1.0 (Hex package)
```

aaronrenner:
@mhanberg I just wanted to let you know I haven't had time to look through the main part of this PR yet, but I plan to pull it down and try it out later this afternoon. Thanks for all of your work on this! :heart:

mhanberg:
@aaronrenner Thank _you_! This PR has been a long time in the making and your recent flurry of activity gave me the motivation to see it through 😄.

michallepicki:
I can take a pass at fixing dialyzer once this is merged,  been doing that at work a lot recently

aaronrenner:
I'm building out a sample app with this using the Readme instructions and have a little initial feedback.

1. The one step I had to add since I'm using ecto, is to update `config/test.exs` to add the following settings
    ```elixir
    config :wallaby,
      otp_app: :wallaby_liveview_playground,
      screenshot_on_failure: true # optional, but nice to have
    ```
2. When a test fails and takes a screenshot, the output looks like this

    Screenshot from 2020-03-11 14-22-32
    Since the feature macro is wrapping the test already, I wonder if we could list the screenshots as part of the error. Something like:
     ```
      1) feature users have names (WallabyLiveviewPlayground.Integration.UserListTest)
         test/wallaby_liveview_playground_web/integration/user_list_test.exs:7
         ** (Wallaby.QueryError) Expected to find 3, visible elements that matched the css '.user' but 0, visible elements were found.
     
         code: |> find(Query.css(".user", count: 3))
         stacktrace:
            (wallaby 0.23.0) lib/wallaby/browser.ex:716: Wallaby.Browser.find/2
           test/wallaby_liveview_playground_web/integration/user_list_test.exs:14: (test)
        
        Screenshots:
        * file:////home/aaron/projects/wallaby_liveview_playground/screenshots/1583958061920414336_users_have_names(1).png
     ```

Finally, this has pretty tight phoenix/ecto integration, so I'm wondering if we should have a sample phoenix app embedded in our test suite. This way we could test this code against a real ecto sandbox and real liveview code. I know I've seen this kind of thing in other libraries and can do some research into how to do it.

mhanberg:
@aaronrenner 

1)
Is that feedback? I have that documented in the Feature module. I can add it to the README as well.

2) I like the screenshots idea! I can try and add that.

aaronrenner:
@mhanberg Yeah, that was the only issue I ran into when following the instructions in the readme. Other than that, it was a super smooth setup.

mix.exs
```diff
@@ -61,8 +61,8 @@ defmodule Wallaby.Mixfile do
       {:credo, "~> 0.9", only: [:dev, :test], runtime: false},
       {:bypass, "~> 1.0.0", only: :test},
       {:excoveralls, "~> 0.7", only: :test},
```
aaronrenner: Does ecto_sql need to be listed as an optional dependency? That might take care of some of those dialyzer errors as well...

michallepicki: Do we still have many dialyzer issues? Do they show up in the library or in tests that are using wallaby?

I just tried running dialyzer on this branch. It only found one issue in one of the tests where `:ok` return type should have been `true`

mhanberg: It would be if we add some phoenix/ecto integration tests.

aaronrenner:
I've been thinking about this for a while and I think this provides some very nice features (screenshots on failure, automatic connection checkout) while being an API we should be able to support in the long term. There are a few long-term improvements

aaronrenner:
Sorry, I didn't mean to close this PR. I was drafting a response and must have clicked the wrong button. 😅

lib/wallaby/feature.ex
```diff
@@ -0,0 +1,216 @@
...
+
+      use Wallaby.DSL
+      import Wallaby.Feature
+
+      setup context do
+        metadata = configure_ecto(context[:async])
+
+        start_session_opts =
+          [metadata: metadata]
+          |> put_create_session_fn(context[:create_session_fn])
```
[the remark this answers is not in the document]
mhanberg: This was added for testing purposes. I'm open to removing it later

integration_test/cases/feature/use_feature_test.exs
```diff
@@ -0,0 +1,37 @@
...
+
+  @expected_capabilities Map.put(
+                           Wallaby.driver().default_capabilities(),
+                           :test,
+                           "I'm a capability"
+                         )
+  @sessions [[capabilities: @expected_capabilities]]
+  feature "reads capabilities from session attribute", %{session: %{capabilities: capabilities}} do
+    assert capabilities.test == @expected_capabilities.test
+  end
```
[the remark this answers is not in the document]
mhanberg: I added it for testing.

mhanberg:
This is ready for a final review.

I was able to fix the broken tests and cleaned up the screenshot log output.

cc/ @aaronrenner @michallepicki

lib/wallaby/feature.ex
```diff
@@ -0,0 +1,220 @@
+defmodule Wallaby.Feature do
+  @includes_ecto Code.ensure_loaded?(Ecto.Adapters.SQL.Sandbox) &&
+                   Code.ensure_loaded?(Phoenix.Ecto.SQL.Sandbox)
```
aaronrenner, line 3: What do you think about moving this below the `@moduledoc`? Generally I've seen moduledoc as the first thing and instance variables below.

mhanberg: Will do 👍

mhanberg:
@aaronrenner Thanks for your patience and support in reviewing this feature! 😄

I'm really excited to get this out there as I think it will greatly improve the DX with Wallaby.

PR #426, opened 2019-04-30
Update ex_doc to 0.20.2 to get the versions dropdown on hexdocs

michallepicki:
I had to bump elixir requirement to 1.7 because of this error when generating docs:

```
** (ExDoc.Retriever.Error) ExDoc 0.19+ requires Elixir v1.7 and later. For earlier Elixir versions, make sure to depend on {:ex_doc, "~> 0.18.0"}
    (ex_doc) lib/ex_doc/retriever.ex:84: ExDoc.Retriever.docs_chunk/1
    (ex_doc) lib/ex_doc/retriever.ex:63: ExDoc.Retriever.get_module/2
    (elixir) lib/enum.ex:1003: Enum.flat_map_list/2
    (ex_doc) lib/ex_doc/retriever.ex:43: ExDoc.Retriever.docs_from_modules/2
    (ex_doc) lib/ex_doc.ex:30: ExDoc.generate_docs/3
    (ex_doc) lib/mix/tasks/docs.ex:298: anonymous fn/6 in Mix.Tasks.Docs.run/3
    (elixir) lib/enum.ex:1826: Enum."-reduce/3-lists^foldl/2-0-"/3
    (ex_doc) lib/mix/tasks/docs.ex:297: Mix.Tasks.Docs.run/3
```

PR #450, opened 2019-08-09
Lower required elixir version; Clean up deps; Bump ex_doc dependency;

michallepicki:
In #426 I unnecessarily increased the required elixir version from 1.3 to 1.7. The reason I did that was because I was unable to generate documentation. Now I know that the issue was caused by an old elixir version set in `.tool-versions`, not the elixir version requirement.

No one complained yet, but I think if we released a new version of Wallaby, people would complain.

I lowered the version to 1.5 - I didn't check but I don't see why Wallaby wouldn't work with it.

Also I upgraded ex_doc once again so we get new cool keyboard shortcuts, I removed `earmark` beause it's a dep to `ex_doc`, changed `ex_doc` env to `docs` (so now to generate docs it's necessary to run it with `MIX_ENV=docs mix docs`) and reordered deps so users can see easily what's added in all deps, dev, test or docs.


mix.exs
```diff
@@ -13,7 +13,7 @@ defmodule Wallaby.Mixfile do
   def project do
     [app: :wallaby,
      version: @version,
-     elixir: "~> 1.7",
+     elixir: "~> 1.5",
```
PragTob: I'd highly suggest to add all now supported elixir versions (minor not all patch levels) back to the travis matrix so that we make sure that we don't break compatibility.

michallepicki: Makes sense, on it!

michallepicki:
Looks like there's one issue with Elixir 1.5.3:
```
$ mix coveralls.safe_travis

Excluding tags: [pending: true]

...........................** (CompileError) test/wallaby/phantom/log_store_test.exs:13: undefined function start_supervised!/1

    (stdlib) lists.erl:1338: :lists.foreach/2

    (stdlib) erl_eval.erl:670: :erl_eval.do_apply/6

    (elixir) lib/code.ex:376: Code.require_file/2

    (elixir) lib/kernel/parallel_require.ex:59: anonymous fn/2 in Kernel.ParallelRequire.spawn_requires/5

The command "mix coveralls.safe_travis" exited with 1.
```
looking into it

.travis.yml
```diff
@@ -11,11 +15,17 @@ env:
...
   - WALLABY_DRIVER=selenium
 matrix:
-  exclude:
+  include:
+  - elixir: 1.5.3
+    otp_release: 20.3
+  - elixir: 1.6.6
+    otp_release: 20.3
+  - elixir: 1.7.4
+    otp_release: 21.3
```
PragTob: If you want to do `include` (which imo is a good idea!) I believe you shouldn't specify them further up under `elixir` and `otp_release` - not sure how that works together with our env variables but it's worth a shot.

Also: 1.5 should work with OTP 18, 1.6 with OTP 19 so that might be good to add to test that compatibility until we know they don't work (iirc some newer dialyzer or credo version might need OTP 19+ or even OTP 20+)

PragTob:
Left a hint on how I believe we could make include work but I'm a bit unsure about env + include but I hope that it works the way I expect it to :)

If you wanted I could also give include a shot - however the earliest I'd probably get to it would be Sunday evening.

.travis.yml
michallepicki: It didn't work with our env variables really, it ran each env variable in default elixir container and then each elixir and otp included without env variables. I removed it and added all options with env variables to the matrix, it will get cleaned up a bit when we deprecate phantomjs.

I added OTP 18 and 19, we'll see if tests pass

michallepicki:
Does anyone have experience with working around travis "double builds" ? My suggestion is to limit "Build pushed pull requests" to just forks, maybe this would work?
```
if: type = push OR (type = pull_request AND fork = true)
```

mhanberg:
> Does anyone have experience with working around travis "double builds" ? My suggestion is to limit "Build pushed pull requests" to just forks, maybe this would work?
> 
> ```
> if: type = push OR (type = pull_request AND fork = true)
> ```
> 
> for some reason after I added this, travis didn't run the builds at all... I reverted this

https://stackoverflow.com/questions/31882306/how-to-configure-travis-ci-to-build-pull-requests-merges-to-master-w-o-redunda#31882307

michallepicki:
@mhanberg This would disable push builds on all other branches. I'm fine with that, @PragTob @keathley ?

PR #459, opened 2019-08-16
Send keys test

jangromko:
Added test for sending text to active element with `send_keys/2`.

Enabled only for Chromedriver, because new versions of Selenium don't support this action. Test should be also useful for implementing W3C webdriver protocol.

### DIFF
--- a/integration_test/cases/browser/send_text_to_active_element_test.exs
+++ b/integration_test/cases/browser/send_text_to_active_element_test.exs
@@ -1,0 +1,24 @@
+defmodule Wallaby.Integration.Browser.SendTextToActiveElementTest do
+  use Wallaby.Integration.SessionCase, async: true
+
+  setup %{session: session} do
+    page = visit(session, "forms.html")
+    {:ok, %{page: page}}
+  end
+
+  describe "send_keys/2" do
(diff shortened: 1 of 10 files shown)


integration_test/cases/browser/send_text_to_active_element_test.exs
```diff
@@ -0,0 +1,24 @@
+defmodule Wallaby.Integration.Browser.SendTextToActiveElementTest do
```
michallepicki: Can we rename it to `send_keys` from `send_text`? This one and the other existing test in browser, not sure why it was `send_text`

jangromko: OK, sure!

jangromko: Done. :)

michallepicki: Awesome, can you also rename `integration_test/cases/browser/send_text_test.exs` to `integration_test/cases/browser/send_keys_test.exs` while at it?

jangromko: Also done, @michallepicki

Issue #365, opened 2018-04-20
(RuntimeError) Wallaby had an internal issue with HTTPoison


stephaniewilkinson:
## Issue

When I switch to using Chrome, I get this error:

```
     ** (RuntimeError) Wallaby had an internal issue with HTTPoison
     stacktrace:
       (wallaby) lib/wallaby/httpclient.ex:32: Wallaby.HTTPClient.make_request/4
       (wallaby) lib/wallaby/experimental/chrome.ex:64: Wallaby.Experimental.Chrome.start_session/1
       (wallaby) lib/wallaby.ex:80: Wallaby.start_session/1
       (school) test/support/feature_case.ex:25: SchoolWeb.FeatureCase.__ex_unit_setup_0/1
       (school) test/support/feature_case.ex:1: SchoolWeb.FeatureCase.__ex_unit__/2
       test/school_web/features/view_test.exs:1: SchoolWeb.UserListTest.__ex_unit__/2
```

Versions:
`elixir: "~> 1.4",
`

## Test Code & HTML

Here's my test:

```
defmodule SchoolWeb.UserListTest do
  use SchoolWeb.FeatureCase, async: true

  import Wallaby.Query, only: [css: 2]

  test "homepage is accessible", %{session: session} do
    session
    |> visit("/")
    |> assert_has(css(".header", count: 1, text: "Welcome"))
  end
end
```

Here's my FeatureCase:
```
defmodule SchoolWeb.FeatureCase do
  use ExUnit.CaseTemplate

  using do
    quote do
      use Wallaby.DSL

      alias School.Repo
      import Ecto
      import Ecto.Changeset
      import Ecto.Query

      import SchoolWeb.Router.Helpers
    end
  end

  setup tags do
    :ok = Ecto.Adapters.SQL.Sandbox.checkout(School.Repo)

    unless tags[:async] do
      Ecto.Adapters.SQL.Sandbox.mode(School.Repo, {:shared, self()})
    end

    metadata = Phoenix.Ecto.SQL.Sandbox.metadata_for(School.Repo, self())
    {:ok, session} = Wallaby.start_session(metadata: metadata)
    {:ok, session: session}
  end
end
```

Here's my HTML:

`  <h1 class="header center">Welcome to School!</h1>
`
Thanks for any help you can give!

keathley:
Thanks for the report @stephaniewilkinson. Can you confirm what version of chrome and chromedriver you have installed?

[...]

PR #463, opened 2019-09-04
Tune Hackney to not overload WebDrivers

mhanberg:
Hopefully this resolves/helps #365 

I have tested this locally with hundreds to thousands of tests running in parallel, and the only reason I've seen it fall apart is because of my computer running out of open file handles.

### DIFF
--- a/lib/wallaby.ex
+++ b/lib/wallaby.ex
@@ -32,7 +32,7 @@ defmodule Wallaby do
     children = [
       supervisor(Wallaby.Driver.ProcessWorkspace.ServerSupervisor, []),
       supervisor(driver(), [[name: Wallaby.Driver.Supervisor]]),
-      :hackney_pool.child_spec(:wallaby_pool, [timeout: 15000, max_connections: 4]),
+      :hackney_pool.child_spec(:wallaby_pool, [timeout: 15_000, max_connections: 4]),
       worker(Wallaby.SessionStore, []),
     ]
 
--- a/lib/wallaby.ex
+++ b/lib/wallaby.ex
@@ -32,6 +32,7 @@ defmodule Wallaby do
     children = [
       supervisor(Wallaby.Driver.ProcessWorkspace.ServerSupervisor, []),
       supervisor(driver(), [[name: Wallaby.Driver.Supervisor]]),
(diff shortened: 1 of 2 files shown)


michallepicki:
> I have tested this locally with hundreds to thousands of tests running in parallel, and the only reason I've seen it fall apart is because of my computer running out of open file handles.

Sounds good, but are you able to reproduce the error from #365 when running without that change?

lib/wallaby.ex
```diff
@@ -32,6 +32,7 @@ defmodule Wallaby do
     children = [
       supervisor(Wallaby.Driver.ProcessWorkspace.ServerSupervisor, []),
       supervisor(driver(), [[name: Wallaby.Driver.Supervisor]]),
+      :hackney_pool.child_spec(:wallaby_pool, [timeout: 15_000, max_connections: 4]),
```
michallepicki: I could not find docs on that in hackney, what is this timeout for? Is it a timeout for starting a worker or does it set default http request timeout for a worker?

keathley: The timeout here is how long the connection will be kept alive in the pool.

michallepicki: @keathley Doesn't it prevent the other configuration from being applied then? Sorry if asking stupid questions, never had to tweak hackney
```
config :wallaby,
  hackney_options: [timeout: :infinity, recv_timeout: :infinity]
```

keathley: No worries :). Both of those options are for each individual request. The timeout option that is being added in this PR sets how long the TCP connection should be kept alive in-between requests.

mhanberg:
> > I have tested this locally with hundreds to thousands of tests running in parallel, and the only reason I've seen it fall apart is because of my computer running out of open file handles.
> 
> Sounds good, but are you able to reproduce the error from #365 when running without that change?

Yes 👍

michallepicki:
> > > I have tested this locally with hundreds to thousands of tests running in parallel, and the only reason I've seen it fall apart is because of my computer running out of open file handles.
> > 
> > 
> > Sounds good, but are you able to reproduce the error from #365 when running without that change?
> 
> Yes +1

Oh god I may be able to finally speed up my 1 hour long test runs

PR #467, opened 2019-09-15
Enhance and standardize capabilities between drivers

### LINKED ISSUE #460: Proposal: Enhance and standardize ability to configure WebDriver capabilities
## Description

For Selenium, you can pass in the capabilities to the `start_session` function to use a custom configuration, but this is not possible using ChromeDriver.

ChromeDriver allows you to configure _some_ of the capabilities by exposing some elixir-level application configuration options.

I propose that we implement a consistent API that allows the user to pass in whatever capabilities they want, while also providing a default set of capabilities to allow a zero-configuration starting point.

This allows for more flexibility as well as opens up the possibilities that we can do with a built in FeatureCase (work for that has been started in #447).

### First-level

I propose that there is a default capabilities configuration that allows developers to get started easily using any of the drivers without needing to configure anything.

### Second-level

Second-level capabilities will be configured using application configuration and will supersede the first-level.

This will replace the configuration options used by ChromeDriver in an attempt to remain consistent with the other drivers as well as provide a transparent API that is more intuitive.

The configuration key will be the driver, so either ChromeDriver or Selenium. Since there is already a configuration option for chromedriver path, these will be moved to inside the new `:chromedriver` option.

The old option to specify the specific Chrome binary will be removed because it can be specified inside the capabilities (this is what Wallaby is doing internally).

#### Example

```elixir
config :wallaby,
  chromedriver: [ 
    path: "path/to/chromedriver",
    capabilities: %{
      # some capabilities
      chromeOptions: %{
        binary: "path/to/googlechrome"
      }
    }
  ],
  selenium: [ 
    capabilities: %{
      # some capabilities
    }
  ]
```

### Third-level

Third-level capabilities will be passed into the `start_session` function and will supersede the second-level capabilities.

#### Example

```elixir
Wallaby.start_session(
  capabilities: %{
    # some capabilities
  }
)
```

---

Let me know what you think!

cc/ @michallepicki

mhanberg:
- Enables the ability to set capabilities by passing them as an option and using application configuration.
- Implements default capabilities for Selenium.
- Moves configuration options for using chrome headlessly, the chrome binary, and the chromedriver binary to the `:chromedriver` key in the `:wallaby` application config.
- Moves documentation for Chrome and Selenium from the README and into their respective moduledocs.
- Raise when we receive a generic error from the driver
- Disables a few credo rules that I don't think are valuable: ParenthesesOnZeroArtiyDefs conflicts with the formatter and PipeChainStart is not useful as a code quality metric

cc/ @michallepicki 

Resolves #460

### DIFF
--- a/.credo.exs
+++ b/.credo.exs
@@ -81,7 +81,7 @@
         {Credo.Check.Readability.ModuleAttributeNames},
         {Credo.Check.Readability.ModuleDoc},
         {Credo.Check.Readability.ModuleNames},
-        {Credo.Check.Readability.ParenthesesOnZeroArityDefs},
+        {Credo.Check.Readability.ParenthesesOnZeroArityDefs, false},
         {Credo.Check.Readability.ParenthesesInCondition},
         {Credo.Check.Readability.PredicateFunctionNames},
         {Credo.Check.Readability.PreferImplicitTry},
@@ -102,7 +102,7 @@
         {Credo.Check.Refactor.NegatedConditionsInUnless},
         {Credo.Check.Refactor.NegatedConditionsWithElse},
         {Credo.Check.Refactor.Nesting},
-        {Credo.Check.Refactor.PipeChainStart},
+        {Credo.Check.Refactor.PipeChainStart, false},
         {Credo.Check.Refactor.UnlessWithElse},
 
         {Credo.Check.Warning.BoolOperationOnSameValues},
--- a/.travis.yml
+++ b/.travis.yml
@@ -61,8 +61,6 @@ matrix:
   - elixir: 1.9.1
     otp_release: 22.0.7
     env: WALLABY_DRIVER=selenium
-# Run in bigger container so we don't run out of memory generating the plt
-sudo: required
 cache:
   directories:
     - _build
@@ -70,18 +68,16 @@ cache:
 
 dist: trusty
 
+addons:
+  chrome: stable
+  firefox: latest
+
 before_script:
-  - export DISPLAY=:99.0
-  - sh -e /etc/init.d/xvfb start
   - export CHROMEDRIVER_VERSION=`curl -s http://chromedriver.storage.googleapis.com/LATEST_RELEASE`
   - curl -L -O "http://chromedriver.storage.googleapis.com/${CHROMEDRIVER_VERSION}/chromedriver_linux64.zip"
   - unzip chromedriver_linux64.zip
   - sudo chmod +x chromedriver
   - sudo mv chromedriver /usr/local/bin
-  - sudo add-apt-repository ppa:ubuntu-toolchain-r/test --yes
-  - sudo apt-get update
-  - sudo apt-get --only-upgrade install google-chrome-stable
-  - sudo apt-get install libstdc++6-4.7-dev
   - MIX_ENV=test mix compile --warnings-as-errors
   - if [ -z ${WALLABY_DRIVER} ]; then travis_wait mix dialyzer --plt; fi # only run dialyzer for the env without drivers
   - bash $TRAVIS_BUILD_DIR/test/tools/start_webdriver.sh
--- a/.travis.yml
+++ b/.travis.yml
@@ -69,6 +69,12 @@ cache:
 dist: trusty
 
 addons:
+  apt:
+    sources:
+      - google-chrome
+    packages:
+      - dpkg
+      - google-chrome-stable
   chrome: stable
   firefox: latest
 
--- a/README.md
+++ b/README.md
@@ -5,9 +5,9 @@
 [![Coverage Status](https://coveralls.io/repos/github/keathley/wallaby/badge.svg?branch=master)](https://coveralls.io/github/keathley/wallaby?branch=master)
 [![Inline docs](http://inch-ci.org/github/keathley/wallaby.svg)](http://inch-ci.org/github/keathley/wallaby)
 
-Wallaby helps you test your web applications by simulating realistic user
-interactions. By default it runs each test case concurrently and manages
-browsers for you. Here's an example test for a simple Todo application:
+Wallaby helps you test your web applications by simulating realistic user interactions.
+By default it runs each test case concurrently and manages browsers for you.
+Here's an example test for a simple Todo application:
 
 ```elixir
 defmodule MyApp.Features.TodoTest do
@@ -26,8 +26,7 @@ defmodule MyApp.Features.TodoTest do
 end
 ```
 
-Because Wallaby manages multiple browsers for you, its possible to test several
-users interacting with a page simultaneously.
+Because Wallaby manages multiple browsers for you, its possible to test several users interacting with a page simultaneously.
 
 ```elixir
 defmodule MyApp.Features.MultipleUsersTest do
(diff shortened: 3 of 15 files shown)


mhanberg:
There was something wrong with Travis and Google Chrome so I had to make some changes to the config.

.credo.exs
```diff
@@ -81,7 +81,7 @@
         {Credo.Check.Readability.ModuleAttributeNames},
         {Credo.Check.Readability.ModuleDoc},
         {Credo.Check.Readability.ModuleNames},
-        {Credo.Check.Readability.ParenthesesOnZeroArityDefs},
+        {Credo.Check.Readability.ParenthesesOnZeroArityDefs, false},
```
PragTob: I think `mix format` enforces this and we should start using mix formatter in this project at some point in time but probably after this is merged in :)

README.md
```diff
@@ -569,64 +553,22 @@ See below for more information on the experimental drivers.
...
-If `chromedriver` is on your `PATH`, then you can skip this step.
-Otherwise (e.g., on NPM-installed `chromedriver` binaries), you can override the path like so:
-
-```elixir
-config :wallaby, chromedriver: "<path/to/chromedriver>"
-```
+Both of these drivers are still "experimental" because they don't support the full API yet and because the implementation is changing rapidly.
+But, if you would like to use them in your project here's what you'll need to do.
 
+Please refer to the [documentation](https://hexdocs.pm/wallaby/Wallaby.Experimental.Chrome.html#content) for further information about using the Chrome driver.
```
PragTob: Small problem: If this is merged but not yet released the docs won't be there. Not sure what a good way to do this is. We can keep it at first, make sub md files or what not :thinking:

.credo.exs
mhanberg: I've been formatting the files as I touch them. 

There isn't a super good way to do this but i dislike having a "format the whole project" commit because it makes git blame useless sometimes.

README.md
mhanberg: I think it is fine.

Some projects will specify at the top of the readme that the readme reflects what is on master, and has a link to the readme at the latest commit.

mhanberg:
@PragTob Thanks for the review!

.credo.exs
PragTob: I mean not useless you can still git blame past that commit.

I think having a whole "mix format this" commit is the best, while not ideal. It allows you to enable the check in CI and the longer you work without it the more files will have that git blame problem.

Also, formatting while touching the files has the same problem (- the functions that were actually changed) because then a change is there but it's also not related to the commit. It's just more dispersed and harder to spot.

README.md
PragTob: I know, not a fan of that as it also requires having that notice on the README and not elsewhere kinda (to be good) :) I also believe that even with that most people will just use the README on master.

One can't protect from it 100% but for big changes where a bigger part of the documentation becomes unavailable I like to be cautious.

.credo.exs
mhanberg: Probably just need to bite the bullet at some point.

README.md
mhanberg: I can add a note here that says this is unreleased and then remove the note when I do the next release.

PragTob: could also link to the code on master that has the docs. 

It's not super needed imo but I think it'd be nice :)

Or a bigger not (like you said before) at the top of the README noting that configuration has changed in a major way linking to the released README version. Might be the best of all words, big note on top, but only if there were big changes :man_shrugging:

.credo.exs
keathley: Contrarily, we could add a .formatter.exs file with no input files specified which should disable the formatter assuming that people's editors aren't doing something shitty and ignoring the .formatter file.

.travis.yml
```diff
@@ -61,27 +61,29 @@ matrix:
...
 
+addons:
+  apt:
+    sources:
+      - google-chrome
+    packages:
+      - dpkg
+      - google-chrome-stable
+  chrome: stable
+  firefox: latest
```
michallepicki: Any reason why you chose chrome stable but firefox latest? I think using firefox stable may be less problematic

README.md
```diff
@@ -569,64 +553,22 @@ See below for more information on the experimental drivers.
...
-config :wallaby,
-  chrome: [
-    binary: "path/to/google/chrome"
-  ]
-```
+In order to run headless chrome you'll need to have ChromeDriver >= 2.30 and chrome >= 60.
+Previous versions of both of these tools _may_ work, but several features will be buggy.
+If you want to setup chrome in a CI environment then you'll still need to install and run xvfb.
+This is due to a bug in ChromeDriver 2.30 that inhibits ChromeDriver from handling input text correctly.
+The bug should be fixed in ChromeDriver 2.31.
```
michallepicki: Do we need this whole comment block at all? I think chrome 60 and chromedriver 2.30 are ancient at this point and we can get rid of this warning.

integration_test/chrome/capabilities_test.exs
```diff
@@ -0,0 +1,149 @@
...
+      expected_capabilities = %{
+        javascriptEnabled: false,
+        loadImages: false,
+        version: "",
+        rotatable: false,
+        takesScreenshot: true,
+        cssSelectorsEnabled: true,
+        nativeEvents: false,
+        platform: "ANY",
+        unhandledPromptBehavior: "accept",
```
michallepicki: At a glance, some of these options don't seem to be supported in Chromedriver anymore. It would be good to double-check each one of them and remove the unnecessary ones some day (not necessarily in this PR).

lib/wallaby/experimental/selenium.ex
```diff
@@ -1,40 +1,102 @@
...
+  @spec start_session([start_session_opts]) :: Wallaby.Driver.on_start_session() | no_return
   def start_session(opts \\ []) do
     base_url = Keyword.get(opts, :remote_url, "http://localhost:4444/wd/hub/")
-    capabilities = Keyword.get(opts, :capabilities, %{})
-    create_session_fn = Keyword.get(opts, :create_session_fn,
-                                    &WebdriverClient.create_session/2)
+    create_session_fn = Keyword.get(opts, :create_session_fn, &WebdriverClient.create_session/2)
 
-    capabilities = Map.merge(default_capabilities(), capabilities)
+    capabilities = Keyword.get(opts, :capabilities, capabilities_from_config())
```
michallepicki: Awesome, thank you! So first we try to get capabilities from opts, then from application config, then take the default ones. This is great!

.travis.yml
michallepicki: OK I can see that we've been using latest before by downloading it manually. I think this may be a leftover from times when selenium webdriver wasn't well supported in Firefox but that's fine.

README.md
michallepicki: I see we still have some code in chromedriver to support these versions... We can remove it and this warning later

.travis.yml
mhanberg: Latest is the terminology it uses for the current version of firefox. 

https://docs.travis-ci.com/user/firefox/#version-aliases

PR #482, opened 2019-11-22
Upgrade bypass

aaronrenner:
This upgrades bypass to 1.0.0, so we can take advantage of the new [`Bypass.stub/4`](https://hexdocs.pm/bypass/1.0.0/Bypass.html#stub/4) function. This will allow us to remove our custom test helpers and make tests easier to read. Instead of having to use the complex `stub_backend` helper:
https://github.com/elixir-wallaby/wallaby/blob/a04118bf33e433b16c3778ee11a171a2e3822e39/test/wallaby/phantom/driver_test.exs#L430-L439

we'll be able to stub only the endpoints we're using and then use bypass like normal.
```elixir
bypass
|> stub_fetch_logs_endpoint(session)
|> Bypass.expect("GET", "/session/#{session.id}/element/#{element.id}/size", fn conn ->
  send_resp(conn, 200, ~s<{
            "sessionId": "#{session.id}",
            "status": 0,
            "value": "not quite sure"
   }>)
end)
```

### Additional notes

I had to bump the minimum tested OTP version to 19, because ranch 1.7 has [dropped support for OTP 18](https://github.com/ninenines/ranch/blob/master/doc/src/guide/migrating_from_1.6.asciidoc). I was able to restrict the versions of `ranch` and `cowlib` to get our tests passing on OTP 18, but I wasn't able to get rid of the deprecation warnings in the test ouput. I figure if we're considering dropping support for Elixir 1.5.x,  then it's fine if we don't jump though hoops to support OTP 18 if Elixir 1.6's min OTP version is 19.


mhanberg:
Looks good

PR #486, opened 2019-11-24
CI adjustments

aaronrenner:
The new Github actions based workflow is great! ❤️  🎉 

Just a couple adjustments...

1. Changed the script to build on all pushes and PRs. I found I needed this change when I was pushing code to a branch on my own fork and it wasn't doing anything.
2. Disabled `fail_fast`. Without this, if one job in the matrix fails, it cancels the rest of them. However, when the rest of the jobs are cancelled, we can't see the test results for the other elixir versions.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,8 +1,5 @@
 name: CI
-on: 
-  pull_request:
-  push:
-    branches: master
+on: [push, pull_request]
 
 jobs:
   linting:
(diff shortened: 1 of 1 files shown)


mhanberg:
@aaronrenner I had set to only build the master branch because I am fairly certain that it will prevent the double builds that we were seeing with Travis.

I might be able to get the "push" checks to not be required for PRs.

PR #490, opened 2019-12-05
Fix link to Chrome driver documentation

jaimeiniesta:


PragTob:
Thank you! :tada:

mhanberg:
I am not sure that this change is the correct fix.

There should be documentation at the original link, but I am seeing that it is not there right now...

PragTob:
@mhanberg yeah I checked and was confused that the original documentation wasn't there. Sorry, didn't dig deeper.

PR #496, opened 2019-12-15
Make Chromedriver escript Compatible

### LINKED ISSUE #495: Make Chromedriver escript Compatible
## Issue
Similar to #192 but for chromedriver. When embedding Wallaby in an escript `Wallaby.Experimental.Chrome.Chromedriver` fails to start.

```
00:13:35.372 [info]  Application wallaby exited: Wallaby.start(:normal, []) returned an error: shutdown: failed to start child: Wallaby.Experimental.Chrome
    ** (EXIT) shutdown: failed to start child: Wallaby.Experimental.Chrome.Chromedriver
            ** (ArgumentError) unknown application: :wallaby
                (elixir) lib/application.ex:699: Application.app_dir/1
                (wallaby) lib/wallaby/experimental/chrome/chromedriver.ex:50: Wallaby.Experimental.Chrome.Chromedriver.wrapper_script/0
                (wallaby) lib/wallaby/experimental/chrome/chromedriver.ex:43: Wallaby.Experimental.Chrome.Chromedriver.start_chromedriver/1
                (wallaby) lib/wallaby/experimental/chrome/chromedriver.ex:23: Wallaby.Experimental.Chrome.Chromedriver.init/1
                (stdlib) gen_server.erl:374: :gen_server.init_it/2
                (stdlib) gen_server.erl:342: :gen_server.init_it/6
                (stdlib) proc_lib.erl:249: :proc_lib.init_p_do_apply/3
```

This is because we can't access files within `priv` in escripts. The fix is similar to #193 where we load the script as a module attribute at compilation.

alecho:
Thanks to @aaronrenner for the solution in #193 which is being used again here.

Fixes #495

### DIFF
--- a/lib/wallaby/experimental/chrome/chromedriver.ex
+++ b/lib/wallaby/experimental/chrome/chromedriver.ex
@@ -5,6 +5,9 @@ defmodule Wallaby.Experimental.Chrome.Chromedriver do
   alias Wallaby.Driver.Utils
   alias Wallaby.Experimental.Chrome
 
+  @external_resource "priv/run_command.sh"
+  @run_script_contents File.read!("priv/run_command.sh")
+
   def start_link do
     GenServer.start_link(__MODULE__, :ok, name: __MODULE__)
   end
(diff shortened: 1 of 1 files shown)


aaronrenner:
@alecho Thanks so much for the PR! ❤️ 

Unfortunately, I just tested this on Mac OS 10.14.6 and it doesn't seem to close the `chromedriver` OS process when the test suite is finished. You can recreate this issue by:

1. Open Activity monitor and search for ChromeDriver
1. Run `WALLABY_DRIVER=chrome mix test`
1. Either let the test suite finish, or press `Ctrl+C Ctrl+C` to kill it
1. Note that the chromedriver process is still running

This is an issue we've had with phantom as well and may need to revisit it. If you have ideas in this area, we'd certainly appreciate the help.

alecho:
Just checked my machine and I see that behavior as well. I'll take a look into this. The other times I've used `Port` I was able to rely on the OS process responding to STDIN closing.

PR #504, opened 2020-01-25
Use bypass to integration test Selenium.start/end_session

aaronrenner:
This transitions from using function injection to using bypass when testing `Selenium.start_session/2` and `Selenium.end_session/1`. Using bypass provides more flexibility in changing the underlying implementation, without having to change the test suite. 

My goal is to remove the `:create_session_fn` and `:end_session_fn` options in an upcoming PR, since they are only used for testing, prevent dialyzer from checking types and make swapping the underlying implementation more difficult.

Finally, I moved the capabilities test out of integration tests into regular unit tests. Bypass seems like the ideal way to ensure the `start_session` call is sends the appropriate capabilities params, and this style of test doesn't require injecting `:create_session_fn`. I also put the tests that modify the environment in their own module, since they need to be run with `async: false`, allowing me to keep the main tests running with `async: true`.


mhanberg:
Should this also be applied to the chromedriver capabilities test?

aaronrenner:
@mhanberg Yeah, I was planning on submitting a follow up PR that does the same thing for Chromedriver. I figured I'd do one at a time to keep them easier to follow, but I can do both at the same time if you think it's better.

mhanberg:
Doesn't matter to me, I just wanted to make sure I was on the same page.

PR #508, opened 2020-01-27
Remove .x elixir/erlang versions in test matrix

aaronrenner:
According to this issue https://github.com/actions/setup-elixir/issues/8, the `1.9.x` syntax is no longer recommended because it can lead to 404s (an issue I ran into this morning when [testing out elixir 1.10](https://github.com/aaronrenner/wallaby/runs/410870743?check_suite_focus=true#step:3:804)).

Although this approach hard-codes the elixir versions we're using, it's also compatible with the latest elixir versions including pre-releases. This PR doesn't add Elixir 1.10 support because it requires a formatting and dialyzer change, that I thought I'd add in a subsequent PR.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,8 +11,17 @@ jobs:
 
     strategy:
       matrix:
-        otp: [21.x]
-        elixir: [1.7.x, 1.8.x, 1.9.x]
+        elixir: ["1.7.4", "1.8.2", "1.9.4"]
+        include:
+          - elixir: "1.9.4"
+            otp: "21.3.8.9"
+            check_formatting: true
+          - elixir: "1.8.2"
+            otp: "21.3.8.9"
+            check_formatting: false
+          - elixir: "1.7.4"
+            otp: "21.3.8.9"
+            check_formatting: false
       fail-fast: false
 
     steps:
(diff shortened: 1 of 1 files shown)


michallepicki:
I don't think this is an issue... Some .x versions were not available because actions used to use Erlang Solutions repos, now they get elixir from hex. And today's issue was just a regular delay that happens pretty often when Elixir is released - Bob has a lot on his plate and sometimes takes longer to build or even trips over and needs a hand.

I like the usage of .x versions because it's obvious what exact version was used in a build, no matter when it ran

aaronrenner:
@michallepicki Ah, I see. I saw the [readme](https://github.com/actions/setup-elixir) had changed and didn't include the `.x` examples anymore, but I missed this line:

> However, values like 22.x are also accepted, and we attempt to resolve them according to semantic versioning rules.

Sounds like it would be better to wait a little bit and try using `1.10.x` again instead of hardcoding version numbers.

michallepicki:
>  I saw the readme had changed and didn't include the .x examples anymore

Hmm, OK that makes sense. So you're probably right and using just `1.10` is preferred now. Still, both should have the same effect and I think 404s wouldn't be caused by .x

Issue #509, opened 2020-02-08
Need ability to pass custom options to Chromedriver executable


aaronrenner:
In issue #473, it appears we'd be able to get more information if allow developers to pass custom options to on chromedriver startup.

Below are the options needed to get the chromedriver debug info written to a log file
```
--log-level=DEBUG --log-path=/path/to/log/chomedriver.log --port=<port>
```
By starting chromedriver with these options, we're able to get more helpful error output like this:

```
[1581199532.142][DEBUG]: DevTools WebSocket Event: Runtime.exceptionThrown 928486C2388B88595FC1D7B4B910AFC3 {
   "exceptionDetails": {
      "columnNumber": 8,
      "exception": {
         "type": "string",
         "value": "Test Error"
      },
      "exceptionId": 1,
      "executionContextId": 1,
      "lineNumber": 15,
      "stackTrace": {
         "callFrames": [ {
            "columnNumber": 8,
            "functionName": "throwError",
            "lineNumber": 15,
            "scriptId": "4",
            "url": "http://localhost:32929/errors.html"
         } ]
      },
      "text": "Uncaught",
      "url": "http://localhost:32929/errors.html"
   },
   "timestamp": 1.581199532141929e+12
}
```

PR #510, opened 2020-02-08
Add tests for wallaby's behavior when starting chromedriver

aaronrenner:
This adds some initial test coverage around how wallaby starts the chromedriver executable. I figured I should add some test coverage to protect the existing behavior before making changes to address #509.

Since chromedriver is started on wallaby's application startup, these tests 

1. stop wallaby
2. make the necessary environment changes (to use fake script and chromedriver as main driver)
3. starts wallaby

The benefit of the `FakeChromedriverScript` is we can record the arguments that chromedriver is being started and assert that default and user configured options are passed correctly. It also makes it much simpler to test checks like running against an old version of chromedriver.

### DIFF
--- a/test/support/application_control.ex
+++ b/test/support/application_control.ex
@@ -1,0 +1,38 @@
+defmodule Wallaby.TestSupport.ApplicationControl do
+  @moduledoc """
+  Test helpers for starting/stopping wallaby during test setup
+  """
+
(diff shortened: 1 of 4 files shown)


aaronrenner:
I'm realizing I have a little more work to do on these test helpers that generate the fake chromedriver script. I'm going to close this PR and open a new one when I'm done changing things.

PR #511, opened 2020-02-10
Remove custom install of ChromeDriver and Selenium form GH Actions

mhanberg:
These come preinstalled on ubuntu-latest now


mhanberg:
Looks like it doesn't come with geckodriver pre-installed... but does come with Firefox installed.

PR #518, opened 2020-03-10
Fix :econnrefused errors on slow chromedriver startup

aaronrenner:
I've been seeing numerous CI failures where the first few chromedriver integration tests fail because chromedriver isn't started yet: https://github.com/aaronrenner/wallaby/runs/470755493?check_suite_focus=true#step:6:77

With this PR, when `Wallaby.start_session/1` is called, this code will call chromedriver's status endpoint (`http://localhost:9515/status`) until it gets a successful response (or raise an error if chromedriver isn't ready within 5 seconds). Once chromedriver has been found to be ready, all future `Wallaby.start_session/1` calls will skip the status check and immediately send then start_session command.

## Notes

1. I made `Wallaby.Experimental.Chrome.Chromedriver` then main API for interacting with the singular wallaby managed chromedriver instance, and moved the actual logic to open the port and start the server to `Wallaby.Experimental.Chrome.Chromedriver.Server`. This move also lets me test the `Server` module with `async: true` because it's not a named process.
1. If the chromedriver server isn't ready within 10 seconds after startup, an exception is raised and the `Server` is restart by its supervisor. That way we shouldn't run into an issue where we're waiting for forever for a server that won't start.
1. I updated the `Wallaby.Experimental.Chrome.init/1` callback to use the newer style of child specs.
1. The readiness checker uses `WebDriverClient.fetch_session_status/1` for the status check. I used this because parsing JWP and W3C status responses are slightly different and this library already has this logic written.  I can definitely do some work to add the status check to our existing webdriver code, but I also thought this would allow us to start migrating more code over to the new API client and be able to take advantage of debugging tools like request logging and work toward adding W3C support to wallaby. I sill have to publish [web_driver_client](https://github.com/aaronrenner/web_driver_client ) `0.1.0` to hex (which is why this PR is in draft), but I'll be happy to publish it if you're OK with moving forward with this code.

Finally, I think we can use a similar approach to fix some of the intermittent timeout issues with PhantomJS that occasionally prevents wallaby from starting. But that's a future PR. :smile:


PragTob:
I don't have the mental capacity for a review right now, but hank you sooo much Aaron! :green_heart: :tada: 

This is one of the most pestering problems and high level your fix sounds great!

IMG_20180317_110254

lib/wallaby/experimental/chrome/chromedriver/server.ex
```diff
@@ -0,0 +1,153 @@
...
+  end
+
+  @impl true
+  def init({chromedriver_path, opts}) do
+    port_number = Utils.find_available_port()
+
+    startup_timeout = Keyword.get(opts, :startup_timeout, @default_startup_timeout)
+    Process.send_after(self(), :ensure_readiness, startup_timeout)
+
+    send(self(), :start_chromedriver)
```
mhanberg: This might be a good place for `handle_continue`

aaronrenner: I was wondering if I could use `handle_continue` since it requires OTP 21+. What is the minimum OTP version we're require for wallaby? I saw Elixir 1.7 supports back to OTP 19. I'm totally happy to change it if it we are only supporting OTP 21+.

mhanberg: I think we are safe to only support OTP 21. In #447 I also made that trade-off to be able to use `__STACKTRACE__`.

aaronrenner:
@mhanberg That would be great if we could leave it as a git-based dependency the next wallaby release. :+1:  I'll push an update to this PR tomorrow as soon as I merge my outstanding PRs on WebDriverClient.

lib/wallaby/experimental/chrome/chromedriver/server.ex
```diff
@@ -54,27 +54,25 @@ defmodule Wallaby.Experimental.Chrome.Chromedriver.Server do
...
-    send(self(), :start_chromedriver)
-
-    {:ok, %State{port_number: port_number, chromedriver_path: chromedriver_path}}
+    {:ok, %State{chromedriver_path: chromedriver_path}, {:continue, nil}}
   end
 
   @impl true
-  def handle_info(:start_chromedriver, state) do
-    %State{port_number: port_number, chromedriver_path: chromedriver_path} = state
+  def handle_continue(_, state) do
```
mhanberg: I think I would still pass the atom `:start_chromedriver` to the continue callback, just to be explicit.

aaronrenner: Thanks for pointing this out. I thought this callback only ran after `init/1`, but it looks like you can return a continue from `handle_call/3` and `handle_cast/2` as well. Updating...

PR #520, opened 2020-03-12
Fix screenshot file url output

aaronrenner:
When wallaby takes a screenshot (either automatically or with `log: true`) it currently outputs a url like this:

```
Screenshot taken, find it at file:////home/aaron/projects/wallaby_liveview_playground/screenshots/1584041849570197413_users_have_names_in_LiveView(1).png
```
This url has an extra slash (`file:////` instead of `file:///`) which is invalid and makes it so it's not clickable in the terminal.

## Fix

This PR fixes the invalid screenshot urls and adds tests for when  `screenshot_dir` is configured with the following types of directories:
* Absolute paths
* Relative paths
* Home directory based paths
* Paths with spaces in them

## Other notes

The majority of this PR is updates to the screenshots test suite. Here are a few high-level changes to the tests
1. Ensure the various types of directories that can be set for `screenshot_dir` work as expected
2. Ensure the file urls that are output with `log: true` are valid and the files they say that were created actually exist. The private `extract_screenshot_urls/1` function grabs the URLs out of the captured output and `path_from_file_url!/1` parses and extracts the absolute path from the file url.
3. Use `Wallaby.SettingsTestHelpers` and `Wallaby.TestSupport.TestWorkspace` helpers to ensure the application environment and file system are cleaned up after the tests. Previously this wasn't always happening if a test failed.
4. Updated `Wallaby.TestSupport.TestWorkspace.mkdir!/1` to accept a `%{random_string}` placeholder in the optional path param. This placeholder is replaced inside the function and makes it so I can remove the `random_string/0` helper function that I was copying across modules. I borrowed the format of the placeholder from `Ecto.Changeset`'s placeholders in its validation messages.
5. Added a `generate_temporary_path/1` function to `Wallaby.TestSupport.TestWorkspace` that generates a path (without creating the directory) and ensures this directory is removed at the end of the test. I made this because I needed a temporary path, but wanted to ensure wallaby made the directory if it didn't exist.

### DIFF
--- a/lib/wallaby/browser.ex
+++ b/lib/wallaby/browser.ex
@@ -227,10 +227,11 @@ defmodule Wallaby.Browser do
 
     name = opts |> Keyword.get(:name, :erlang.system_time()) |> to_string
     path = path_for_screenshot(name)
-    File.write!(path, image_data)
+
+    write_screenshot!(path, image_data)
 
     if opts[:log] do
-      IO.puts("Screenshot taken, find it at file:///#{path}")
+      IO.puts("Screenshot taken, find it at #{build_file_url(path)}")
     end
 
     Map.update(screenshotable, :screenshots, [], &(&1 ++ [path]))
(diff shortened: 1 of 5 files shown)


aaronrenner:
I'm not quite sure what's going on with the CodeCov report... I added more test coverage, not less.  :thinking:

lib/wallaby/browser.ex
```diff
@@ -1220,11 +1221,23 @@ defmodule Wallaby.Browser do
...
+    expanded_path = Path.expand(path)
+    :ok = expanded_path |> Path.dirname() |> File.mkdir_p!()
+
+    :ok = File.write!(expanded_path, image_data)
+
+    :ok
+  end
+
   defp screenshot_dir do
     Application.get_env(:wallaby, :screenshot_dir) || "#{File.cwd!()}/screenshots"
```
mhanberg: This isn't related to your change, but this should probably be written as

```suggestion
    Application.get_env(:wallaby, :screenshot_dir, "#{File.cwd!()}/screenshots") 
```

PR #521, opened 2020-03-16
Start phantomjs servers in parallel

aaronrenner:
Currently, phantomjs processes are started completely before returning from the init/1 callback. Since supervisors start children one-by-one, this is causing phantomjs servers to be started sequentially, instead of in parallel. 

This PR moves the phantomjs startup process out of `init/1` and adds a `wait_until_ready/1` call that polls the phantomjs process until it's ready (the same thing we do with chromedriver) so the phantomjs servers can be started in parallel. This PR also includes a few other benefits.

1. Currently the phantomjs starting logic was spread across 3 different modules and made it very difficult to understand what's going on. This consolidates the logic into a single `server.ex`, which I think makes it easier to follow.
2. Simplifies `run_phantom.sh` to more closely align with the suggestions in the Elixir [Port](https://hexdocs.pm/elixir/Port.html#module-zombie-operating-system-processes) docs. This also fixes a race condition that was in the previous version of the script.
3. Moves to more integration-style testing to ensure the server actually works than unit tests + mocks.


lib/wallaby/phantom/server.ex
```diff
@@ -2,117 +2,246 @@ defmodule Wallaby.Phantom.Server do
...
+    @moduledoc false
+
+    @type port_number :: non_neg_integer()
+    @type t :: %__MODULE__{
+            port_number: port_number | nil,
+            phantomjs_path: String.t(),
+            phantomjs_user_provided_args: [String.t()],
+            wrapper_script_port: port | nil,
+            wrapper_script_os_pid: pid | nil,
+            phantomjs_os_pid: pid | nil,
```
aaronrenner: Just noticed these typespecs should be `os_pid | nil` instead. Updating...

aaronrenner:
I just found the updated wrapper script wasn't respecting `SIGTERM`, `SIGHUP` and `SIGINT`. I updated the script to handle these and added a couple more checks in our tests to ensure no zombie processes are laying around. If this looks good, I'd like to use this as a model for how we handle and test chromedriver.

PR #527, opened 2020-05-01
Extract poolboy logic into ResourcePool module

aaronrenner:
This wraps the poolboy logic into its own module, making it simpler to implement webdriver server pools. `Wallaby.Phantom.ServerPool` is a good example of how this simplifies implementing a pool.


mhanberg:
What do you think about implementing the pool using José's [nimble_pool](https://hexdocs.pm/nimble_pool/NimblePool.html) library?

PR #538, opened 2020-06-15
Update ServerSupervisor to be a DynamicSupervisor

mhanberg:
Also removes call to `Process.sleep` in the test in favor of recursion with a timeout.

I was investigating some flaky tests that involved the `assert File.exists?(workspace_path)` lines in the tests, and noticed that was a `:simple_one_for_one` supervisor. Converting it to a `DynamicSupervisor` seems to have fixed flakiness.

The call to `Process.sleep` was not part of the flakiness, but I converted it to use recursion with a timeout instead. Let me know what you think about that.

### DIFF
--- a/lib/wallaby/driver/process_workspace/server_supervisor.ex
+++ b/lib/wallaby/driver/process_workspace/server_supervisor.ex
@@ -1,24 +1,26 @@
 defmodule Wallaby.Driver.ProcessWorkspace.ServerSupervisor do
   @moduledoc false
 
-  use Supervisor
+  use DynamicSupervisor
 
   alias Wallaby.Driver.ProcessWorkspace.Server
 
   @spec start_link :: Supervisor.on_start()
   def start_link do
-    Supervisor.start_link(__MODULE__, [], name: __MODULE__)
+    DynamicSupervisor.start_link(__MODULE__, [], name: __MODULE__)
   end
 
-  @spec start_server(pid, String.t()) :: Supervisor.on_start_child()
+  @spec start_server(pid, String.t()) :: DynamicSupervisor.on_start_child()
   def start_server(process_pid, workspace_path) do
-    Supervisor.start_child(__MODULE__, [process_pid, workspace_path])
+    DynamicSupervisor.start_child(__MODULE__, %{
+      id: Server,
+      start: {Server, :start_link, [process_pid, workspace_path]},
+      restart: :transient
+    })
   end
 
-  @impl Supervisor
+  @impl DynamicSupervisor
   def init([]) do
-    children = [worker(Server, [], restart: :transient)]
-
-    supervise(children, strategy: :simple_one_for_one)
+    DynamicSupervisor.init(strategy: :one_for_one)
   end
 end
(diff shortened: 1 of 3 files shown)


lib/wallaby/driver/process_workspace/server_supervisor.ex
```diff
@@ -1,24 +1,26 @@
...
 
-  use Supervisor
+  use DynamicSupervisor
 
   alias Wallaby.Driver.ProcessWorkspace.Server
 
   @spec start_link :: Supervisor.on_start()
   def start_link do
-    Supervisor.start_link(__MODULE__, [], name: __MODULE__)
+    DynamicSupervisor.start_link(__MODULE__, [], name: __MODULE__)
```
aaronrenner, line 10: Lately, if I have a singleton instance of a process like this one, I've been storing the process name in a variable named `@instance`.

```elixir
  @instance __MODULE__

  def start_link do
    DynamicSupervisor.start_link(__MODULE__, [], name: @instance)
  end
```

The benefit I've found is it seems to make subsequent calls to that process easier to understand.

```elixir
# It takes me a minute to figure out what __MODULE__ is here...
def start_server(process_pid, workspace_path) do
  DynamicSupervisor.start_child(__MODULE__, {Server, [process_pid, workspace_path})
end

# versus

# Oh, the first argument to start_child/2 is the supervisor instance
def start_server(process_pid, workspace_path) do
  DynamicSupervisor.start_child(@instance, {Server, [process_pid, workspace_path})
end
```

I know this isn't related to the transition from `Supervisor` to `DynamicSupervisor`, but since I saw it, I thought I'd bring it up and see what you think.

test/wallaby/driver/process_workspace_test.exs
```diff
@@ -14,6 +14,24 @@ defmodule Wallaby.Driver.ProcessWorkspaceTest do
...
+    e ->
+      passed_timeout? = now_in_milliseconds() - start >= timeout
+
+      if passed_timeout? do
+        reraise e, __STACKTRACE__
+      else
+        with_timeout(doer, start, timeout)
+      end
+  end
+
```
aaronrenner, line 34: What do you think about moving these private functions to the bottom of the module? When I first open up the file, it's a lot to digest when I'm really just curious in what the tests are.

lib/wallaby/driver/process_workspace/server_supervisor.ex
```diff
@@ -1,24 +1,26 @@
...
 
-  @spec start_server(pid, String.t()) :: Supervisor.on_start_child()
+  @spec start_server(pid, String.t()) :: DynamicSupervisor.on_start_child()
   def start_server(process_pid, workspace_path) do
-    Supervisor.start_child(__MODULE__, [process_pid, workspace_path])
+    DynamicSupervisor.start_child(__MODULE__, %{
+      id: Server,
+      start: {Server, :start_link, [process_pid, workspace_path]},
+      restart: :transient
+    })
```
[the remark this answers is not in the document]
mhanberg: Sounds good 👍

test/wallaby/driver/process_workspace_test.exs
mhanberg: Sounds good 👍

lib/wallaby/driver/process_workspace/server_supervisor.ex
mhanberg: I'm don't think I find that much clearer, in either case, you need to go to the top of the file to see what the value of it is.

PR #543, opened 2020-06-19
Fix Query type specifications

### LINKED ISSUE #542: Dialyzer error when specifying Queries in module attributes in compiled helpers
Example Repo: https://github.com/TheFirstAvenger/wallaby_dialyzer_issue

If you have compiled test helpers, and use module attributes to specify Queries, then functions that call `fill_in`, `click`, or `assert_has` will throw the Dialyzer "no return" error. These show up as errors in IDE tools like ElixirLS.

```elixir
  @doc """
  This version has inline Queries.
  There is no dialyzer error on this version
  """
  def log_in_no_error(session) do
    session
    |> Browser.visit("/login")
    |> Browser.fill_in(Query.text_field("username"), with: "me")
    |> Browser.click(Query.css("button", text: "Log in"))
    |> Browser.assert_has(Query.css("h1", text: "Logged In!"))
  end

  @username_field Query.text_field("username")
  @log_in_button Query.css("button", text: "Log in")
  @logged_in_h1 Query.css("h1", text: "Logged In!")

  @doc """
  This version uses module atttributes for it's queries
  This results in the Dialyzer error "Function log_in_has_no_return/1 has no local return."
  """
  def log_in_has_no_return(session) do
    session
    |> Browser.visit("/login")
    |> Browser.fill_in(@username_field, with: "me")
    |> Browser.click(@log_in_button)
    |> Browser.assert_has(@logged_in_h1)
  end
```

michallepicki:
Fixes #542 
I narrowed this down to `Query.compile(@query_from_module_attribute)`. Looks like with the query created at compile time, dialyzer has more information to find discrepancies. Types in the query module were missing some possible values - I updated them to be slightly more general and cover more of valid queries. This fixes the reported bugs but there still could be more valid query structs that are not covered by the current types, so we might need to revisit this in the future.

### DIFF
--- a/lib/wallaby/query.ex
+++ b/lib/wallaby/query.ex
@@ -98,24 +98,26 @@ defmodule Wallaby.Query do
           | :option
           | :select
           | :file_field
+          | :attribute
   @type attribute_key_value_pair :: {String.t(), String.t()}
   @type selector ::
           String.t()
-          | :attribute_key_value_pair
+          | attribute_key_value_pair()
   @type html_validation ::
           :bad_label
           | :button_type
           | nil
   @type conditions :: [
-          count: non_neg_integer,
-          text: String.t(),
+          count: integer,
+          minimum: integer,
+          maximum: integer,
+          text: String.t() | nil,
           visible: boolean() | :any,
           selected: boolean() | :any,
-          minimum: non_neg_integer,
-          at: pos_integer
+          at: integer | :all
         ]
   @type result :: list(Element.t())
-  @type opts :: nonempty_list()
+  @type opts :: list()
 
   @type t :: %__MODULE__{
           method: method(),
--- a/lib/wallaby/query.ex
+++ b/lib/wallaby/query.ex
@@ -108,9 +108,9 @@ defmodule Wallaby.Query do
           | :button_type
           | nil
   @type conditions :: [
-          count: integer,
-          minimum: integer,
-          maximum: integer,
+          count: non_neg_integer,
+          minimum: non_neg_integer,
+          maximum: non_neg_integer,
           text: String.t() | nil,
           visible: boolean() | :any,
           selected: boolean() | :any,
--- a/lib/wallaby/query.ex
+++ b/lib/wallaby/query.ex
@@ -114,7 +114,7 @@ defmodule Wallaby.Query do
           text: String.t() | nil,
           visible: boolean() | :any,
           selected: boolean() | :any,
-          at: integer | :all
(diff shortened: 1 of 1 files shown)


michallepicki:
@aaronrenner @mhanberg Should I do anything with codecov? I didn't even touch the tests :thinking:

mhanberg:
It’ll go back up as the tests finish.

lib/wallaby/query.ex
```diff
@@ -98,24 +98,26 @@ defmodule Wallaby.Query do
...
   @type html_validation ::
           :bad_label
           | :button_type
           | nil
   @type conditions :: [
-          count: non_neg_integer,
-          text: String.t(),
+          count: integer,
+          minimum: integer,
+          maximum: integer,
```
aaronrenner, line 113: Is there a case where we'd expect count, minimum, or maximum to be negative?

michallepicki: Oh you're probably right, these should probably stay as `non_neg_integer`. I was changing the `at` to be correct because I believe underneath it will be passed to `Enum.at`, and I unnecessarily changed these three. I'll update!

michallepicki, line 113:
```suggestion
          count: non_neg_integer,
          minimum: non_neg_integer,
          maximum: non_neg_integer,
```

michallepicki: I double checked dialyzer on the repro project locally and it doesn't affect the fix so it should be fine :)

lib/wallaby/query.ex
```diff
@@ -98,24 +98,26 @@ defmodule Wallaby.Query do
...
           count: non_neg_integer,
-          text: String.t(),
+          minimum: non_neg_integer,
+          maximum: non_neg_integer,
+          text: String.t() | nil,
           visible: boolean() | :any,
           selected: boolean() | :any,
-          minimum: non_neg_integer,
-          at: pos_integer
+          at: integer | :all
```
aaronrenner, line 117: Should this also be a `non_neg_integer`?

```suggestion
          at: non_neg_integer | :all
```

Here's the relevant test coverage. It seems like -1 isn't valid...

https://github.com/elixir-wallaby/wallaby/blob/bf59c62493469c40bce7467cab0cc6f5aa0a1289/integration_test/cases/query_test.exs#L122-L145

michallepicki: I would think that `-1` returns the last element, but maybe we don't support it right now. Looks like the test catches that, you're probably right. I'll revert.

Issue #550, opened 2020-07-13
How to list current active sessions ?


r8code:
## Question
Selenium Chrome

Hi , first thank you for this great library 
I want to ask is there any API  to list current active sessions 
I saw there is Wallaby.SessionStore.list_sessions_for   but  if I  had end_session  , the list is not empty 
Is there any way to list current active sessions ? 
Thank you

mhanberg:
There is currently no public API for retrieving the sessions.

Do you mind sharing why you want to get the list of sessions?

r8code:
Hi, I need to retrieving the sessions, to make sure there's no dangling chromedriver port opened by selenium process ,
if I can retrieve the sessions, I can make sure that the current status of selenium is the same as the application
Thanks for the reply

bobwaycott:
I'll chime in here and say I'm also looking for a means of identifying active sessions (and verifying a closed/ended session is truly gone). Use case here is a separate process interacting with webpages in a non-test setting. There could be interaction between the calling process and the `wallaby` process, and want to be able to do something like perform an action using session ID `X`, then when the task is complete, know the session has been truly ended and can't be used again. Noticed the same issue here where I can see all sessions I've ever used in `Wallaby.SessionStore.list_sessions_for/1`, but sessions still appear there after calling `Wallaby.end_session/1`.

mhanberg:
> I saw there is Wallaby.SessionStore.list_sessions_for but if I had end_session , the list is not empty

> Noticed the same issue here where I can see all sessions I've ever used in Wallaby.SessionStore.list_sessions_for/1, but sessions still appear there after calling Wallaby.end_session/1.

I think the case here (ending a session, but not seeing that reflected in the session store) is related to #554.

> Hi, I need to retrieving the sessions, to make sure there's no dangling chromedriver port opened by selenium process ,
if I can retrieve the sessions, I can make sure that the current status of selenium is the same as the application
Thanks for the reply

@r8code I'm not really understanding what you are saying here.

Are you seeing instances of Chrome remaining opening after your test suite finishes?

> I'll chime in here and say I'm also looking for a means of identifying active sessions (and verifying a closed/ended session is truly gone). Use case here is a separate process interacting with webpages in a non-test setting. There could be interaction between the calling process and the wallaby process, and want to be able to do something like perform an action using session ID X, then when the task is complete, know the session has been truly ended and can't be used again.

[...]

PR #558, opened 2020-07-31
Ensure sessions are correctly deleted from ETS by passing the pid along

bobwaycott:
ETS records weren't being correctly removed from `SessionStore`, leaving stale sessions around in `SessionStore.list_sessions_for`. By returning the `pid` from the `:ets.select`, we can ensure a correct match & delete of the session.

Could be useful to add a `defdelegate active_sessions` in the main `Wallaby` interface that calls `SessionStore.list_sessions_for`, but didn't want to make that assumption in this PR.

Fixes the issue pointed out in #550

### DIFF
--- a/lib/wallaby/session_store.ex
+++ b/lib/wallaby/session_store.ex
@@ -39,11 +39,11 @@ defmodule Wallaby.SessionStore do
   def handle_call({:demonitor, session}, _from, _state) do
     result =
       :ets.select(:session_store, [
-        {{{:"$1", :"$2", :"$3"}, :"$4"}, [{:==, :"$2", session.id}], [{{:"$1", :"$3", :"$4"}}]}
-        ])
+        {{{:"$1", :"$2", :"$3"}, :_}, [{:==, :"$2", session.id}], [{{:"$1", :"$3"}}]}
+      ])
 
     case result do
-      [{ref, pid, _session}] ->
+      [{ref, pid}] ->
         true = Process.demonitor(ref)
         :ets.delete(:session_store, {ref, session.id, pid})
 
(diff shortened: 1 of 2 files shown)


mhanberg:
Thanks for the PR! Do you mind adding some tests?

bobwaycott:
There are currently no tests for `SessionStore` in the project. What are your thoughts on how `SessionStore` should be tested?

mhanberg:
It think it's fine to create a new unit test module for the session store, and test that it can store and delete multiple sessions for the same test pid, and remove them both when that test pid goes down.

lib/wallaby/session_store.ex
```diff
@@ -39,13 +39,13 @@ defmodule Wallaby.SessionStore do
...
     result =
       :ets.select(:session_store, [
-        {{{:"$1", :"$2", :"$3"}, :"$4"}, [{:==, :"$2", session.id}], [{{:"$1", :"$4"}}]}
-      ])
+        {{{:"$1", :"$2", :"$3"}, :"$4"}, [{:==, :"$2", session.id}], [{{:"$1", :"$3", :"$4"}}]}
+        ])
 
     case result do
-      [{ref, _}] ->
+      [{ref, pid, _session}] ->
```
mhanberg, line 46: should this be `[{ref, _session, pid}]`?

bobwaycott: I should probably switch the return value order to match the way the data is stored in ETS. This call is based on the way the result is *returned from*, not *stored in* ETS (you can see the `:ets.delete` call matches the storage order.

mhanberg: Oh, I see now that I misread it.

bobwaycott:
@mhanberg added a couple straightforward tests to verify the ETS table adds & correctly removes sessions

bobwaycott:
Happy to help!

bobwaycott:
@mhanberg sorry to comment again on a closed PR, but wanted to check in on what the release cycle is like for wallaby being updated in Hex? Would love to have this fix available, and curious if I should install from github, or if a new release will be made with this fix?

mhanberg:
I'm currently adding some more tests. I'll release a patch release once I'm done with that. You can install from github master until then if you'd like.

PR #559, opened 2020-08-05
Fix typespec issues

bobwaycott:
When using `Wallaby.Browser.has_text?/2`  and `Wallaby.Browser.assert_text/2` in a codebase, dialyzer throws warnings when you call these funcs with a `Wallaby.Session`, even though it is supported. Update the `@spec` declarations to use `parent` instead of `Element.t()`, so dialyzer can properly determine the usage is valid.

Additionally, in `Wallaby.Browser`, there is a `@type locator :: String.t()` def, which is only used 3x in the module, whereas `String.t()` is used 33x, sometimes in sibling `@spec` statements for the same multi-head function. Dropped `locator` and made the 3 instances `String.t()` instead.

### DIFF
--- a/lib/wallaby/browser.ex
+++ b/lib/wallaby/browser.ex
@@ -123,7 +123,6 @@ defmodule Wallaby.Browser do
   @type parent ::
           element
           | session
-  @type locator :: String.t()
   @type opts :: Query.opts()
 
   @default_max_wait_time 3_000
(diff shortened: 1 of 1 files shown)


bobwaycott:
I'd like to mention another dialyzer issue I haven't yet figured out the best way to resolve. 

It seems dialyzer keeps getting confused on various `Browser` funcs (like `click/2`) that spec a return type of `parent`, yet interprets them as always expecting to return an `Element`, even though `parent` specifies it will be either `Element` *or* `Session`.  Any thoughts on how to satisfy dialyzer here?

bobwaycott:
Any feedback here? Would be really nice to see dialyzer happy when using this in other projects.

michallepicki (approved):
This all makes sense to me. Can you update the branch from current master @bobwaycott ?

michallepicki:
Branch updated

PR #564, opened 2020-08-18
Make assert_text pipeable

Jcambass:
This patch changes the `assert_text\2` and the `assert_text\3` methods to return the parent rather than `true` if the assertion passes.

This will allow easy piping of the method.

Closes https://github.com/elixir-wallaby/wallaby/issues/561

### DIFF
--- a/integration_test/cases/browser/assert_text_test.exs
+++ b/integration_test/cases/browser/assert_text_test.exs
@@ -13,7 +13,7 @@ defmodule Wallaby.Integration.Browser.AssertTextTest do
     refute has_text?(element, "rain")
   end
 
-  test "assert_text/2 waits for presence of text and and returns true if found", %{
+  test "assert_text/2 waits for presence of text and and returns the parent if found", %{
     session: session
   } do
     element =
@@ -21,7 +21,7 @@ defmodule Wallaby.Integration.Browser.AssertTextTest do
       |> visit("wait.html")
       |> find(Query.css("#container"))
 
-    assert assert_text(element, "main")
+    assert element == assert_text(element, "main")
   end
 
   test "assert_text/2 will raise an exception for text not found", %{session: session} do
@@ -39,6 +39,6 @@ defmodule Wallaby.Integration.Browser.AssertTextTest do
     session
     |> Browser.visit("wait.html")
 
-    assert Browser.assert_text(session, "main")
+    assert session == Browser.assert_text(session, "main")
   end
 end
--- a/lib/wallaby/browser.ex
+++ b/lib/wallaby/browser.ex
@@ -764,7 +764,25 @@ defmodule Wallaby.Browser do
   end
 
   @doc """
-  Matches the Element's content with the provided text
+  Matches the parent's content with the provided text.
+
+  Returns a boolean that indicates if the text was found.
+
+  ## Examples
+
+  ```
+  session
+  |> visit("/")
+  |> has_text?("Login")
+  ```
+
+  Example providing query:
+
+  ```
+  session
+  |> visit("/")
+  |> has_text?(Query.css(".login-button"), "Login")
+  ```
   """
   @spec has_text?(parent, String.t()) :: boolean()
   @spec has_text?(parent, Query.t(), String.t()) :: boolean()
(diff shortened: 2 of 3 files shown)


integration_test/cases/browser/assert_text_test.exs
```diff
@@ -21,7 +21,7 @@ defmodule Wallaby.Integration.Browser.AssertTextTest do
       |> visit("wait.html")
       |> find(Query.css("#container"))
 
-    assert assert_text(element, "main")
+    assert element = assert_text(element, "main")
```
mhanberg, line 24:
```suggestion
    assert element == assert_text(element, "main")
```

integration_test/cases/browser/assert_text_test.exs
```diff
@@ -39,6 +39,6 @@ defmodule Wallaby.Integration.Browser.AssertTextTest do
     session
     |> Browser.visit("wait.html")
 
-    assert Browser.assert_text(session, "main")
+    assert %Wallaby.Session{} = Browser.assert_text(session, "main")
```
mhanberg, line 42:
```suggestion
    assert session == Browser.assert_text(session, "main")
```

lib/wallaby/browser.ex
```diff
@@ -822,8 +822,8 @@ defmodule Wallaby.Browser do
   @doc """
   Matches the Element's content with the provided text and raises if not found
   """
-  @spec assert_text(Element.t(), String.t()) :: boolean()
-  @spec assert_text(parent, Query.t(), String.t()) :: boolean()
+  @spec assert_text(Element.t(), String.t()) :: parent
+  @spec assert_text(parent, Query.t(), String.t()) :: parent
```
mhanberg, line 826: The docs and specs should be updated to reflect the actual types that are permitted. (I know this isn't related to your change).

`assert_text` and `has_text?` can both take sessions or elements, but the specs and docs don't reflect that.

Can you change the specs and docs to be accurate, as well as well as enhance the docs with an example? It would be very much appreciated.

integration_test/cases/browser/assert_text_test.exs
mhanberg: Can you also update the name of the tests in the file to be accurate?

lib/wallaby/browser.ex
```diff
@@ -783,9 +783,27 @@ defmodule Wallaby.Browser do
   end
 
   @doc """
-  Matches the Element's content with the provided text
+  Matches the Element's content with the provided text.
```
Jcambass, line 786: @mhanberg Is using the word `Element` here ok (I see that it is used in other places) or would `Parent` be more correct here?

mhanberg: "Matches the paren't content with the provided test." is probably okay.

Jcambass: Updated to `parent's`.

Jcambass:
@mhanberg Resolved merge conflicts.

Jcambass:
I see that you closed the PR and that the issue is still open. Let me know if there are some changes I would need to make for it to be mergable or if the issue is also to be closed.

mhanberg:
Looks like when I deleted to master branch, it closed all the PRs 😅.

mhanberg:
I'll get these PRs re-opened in the morning. I think I just need to changed the target branch to main.

mhanberg:
Turns out I can't modify the target branch myself, please update it to main by hitting  "Edit" next to the PR title.

Jcambass:
@mhanberg I just tried but I didn't get the option. I think this might be because the PR is closed.

mhanberg:
I forgot that I can just push up another master branch while i clean these up. Should be good now.

Jcambass:
I'll fix those merge conflicts and then this should be good to go.

mhanberg:
Pleas update the CHANGELOG.md with the change, following the structure for breaking changes, then I can merge it.

mhanberg:
Thanks a lot! Sorry this took so long to get through.

Jcambass:
No worries! Thanks for your assistance :)

PR #567, opened 2020-08-31
Fix Selenium uploads

bobwaycott:
Currently, Wallaby's `Browser.attach_file` uses `set_value` to set a file path as the value of an file input.

Unfortunately, this does not actually attach a file when using Selenium, and submitting a form when using this function results in errors for invalid multipart payloads. This problem is more pronounced when using a remote Selenium instance, as a file input can only attach a file that is available on the local machine.

Selenium's standalone server provides a means of handling this that is implemented in other libraries (e.g., [the official Python implementation](https://github.com/SeleniumHQ/selenium/blob/941dc9c6b2e2aa4f701c1b72be8de03d4b7e996a/py/selenium/webdriver/remote/webelement.py#L511)).

The basic outline of how this is supposed to work with Selenium servers is as follows:

1. Instead of calling `set_value` on a file input, you should call `send_keys` on the file input with the local file path
2. `send_keys` should detect if the `keys` supplied are actually a local file path
3. When a local file is detected:
  a. create a zip file containing the local file
  b. base64 encode the zip file from step 3a
  c. upload encoded zip from 3b to Selenium's `<session_id>/file` endpoint
4. step c will return the path (local to Selenium) where the file was uploaded and attached to the active Selenium session
5. Call `send_keys` on the file input, this time sending the local-to-Selenium path as the value for the file input

Upon submitting the form, your local file will have been successfully uploaded to Selenium and attached to the form for proper submission.

This PR implements correct and successful file uploads for Selenium in `Browser.attach_file`—it detects if the `driver` is Selenium, and follows the path above, otherwise, falls back to `set_value` for maintaining Chrome compatibility (which seems to be fine with `set_value` and properly uploads a file *locally* (but I'm unclear on how you'd do that remotely and verify)).

**NOTE:** All existing tests pass, and much of this functionality relies on modifying the `Selenium` module, since it is Selenium-specific functionality. However, I'm not sure how best to add tests for the remote Selenium server case, as that requires an actual remote Selenium server instance running. This doesn't change any test behavior as it is written, and verifying this with a successful upload would require more in the Wallaby project to process an uploaded file (which is maybe why this hasn't been caught before?).

### DIFF
--- a/lib/wallaby/browser.ex
+++ b/lib/wallaby/browser.ex
@@ -202,8 +202,9 @@ defmodule Wallaby.Browser do
   Attaches a file to a file input. Input elements are looked up by id, label text,
   or name.
 
-  When dealing with a remote Selenium server, the file must be uploaded to the
-  remote Selenium server, else attaching a file fails.
+  When dealing with Selenium server (especially a remote instance), the file
+  must be uploaded to Selenium via `send_keys` recognizing a local file,
+  else attaching a file fails.
   """
   @spec attach_file(parent, Query.t(), path: String.t()) :: parent
   def attach_file(parent, query, path: path) do
--- a/lib/wallaby/browser.ex
+++ b/lib/wallaby/browser.ex
@@ -211,7 +211,7 @@ defmodule Wallaby.Browser do
       true ->
         # local/remote selenium will only properly attach
         # & upload a local file with `send_keys`
-        send_keys(parent, query, [path])
+        send_keys(parent, query, List.wrap(path))
 
       false ->
         # retain existing chrome compatibility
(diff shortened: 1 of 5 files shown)


bobwaycott:
Argh. Looks like I need to do some more work on tests.

bobwaycott:
Hey, so, I'm a bit confused on how to dig into these unexpected `Selenium (1.8.x/21.x)` integration test failures. Can anyone lend a hand or point me in the right direction here for how to proceed? I'm getting no test failures locally and am just not sure how to duplicate that weird batch of failures that are showing currently.

mhanberg:
Thanks!

PR #577, opened 2020-12-02
Switch to httpc and Don't Retry

JEG2:
Hopefully this is better that filling up a hackney pool with failed requests.


JEG2:
Sorry, I didn't mean to post this here.

PR #592, opened 2021-03-10
README.md: Improve instructions about async tests with Phoenix

jonleighton:
To provide example configuration when an application has additional
resources (above just the SQL database) that should be shared between
processes.

See https://github.com/elixir-wallaby/wallaby/discussions/591


mhanberg:
I tested this out to make sure I could follow the instructions and it works great!

https://github.com/mhanberg/wallamox

jonleighton:
Great. I like the name of your repo 🙂

PR #615, opened 2021-08-14
add within_shadow_dom

superchris:
This PR is not ready yet, but I wanted to begin discussion about it. For those of us using custom elements on an Elixir/Phoenix project, having a way to write feature tests that interact with them is essential. This PR adds a `within_shadow_dom` function to do just that. My questions are: Should this go in the Driver behaviour? If not, is there a way to expose driver specific functions (eg only support this in Chrome)?


mhanberg:
Hi!

I'm currently on vacation, but I'll be able to take a look next week. 

If I don't chime in, by next weekend feel free to ping me.

mhanberg:
Plopping this here for posterity: https://groups.google.com/g/ruby-capybara/c/pWJEk7y2NM8/m/-niMWeQTCQAJ

mhanberg:
@superchris I believe shadow root functionality has been added to Chrome 101 (which I think will drop very soon), so it would great if this could be ported to using the builtin endpoints.

- https://w3c.github.io/webdriver/#shadow-root
- https://w3c.github.io/webdriver/#get-element-shadow-root
- https://w3c.github.io/webdriver/#find-elements-from-shadow-root
- https://w3c.github.io/webdriver/#find-element-from-shadow-root

It currently is not available for Firefox 101 or Safari, but I suspect they will come soon.

- https://wpt.fyi/results/webdriver/tests?label=experimental&label=master&aligned

mhanberg:
@superchris I think shadow root support is in the webdrivers now, would you want to take a crack at adding support for it? If not I can take care of this soon.

Basically just need to add functions to hit the various shadow root related endpoints in the webdriver spec https://w3c.github.io/webdriver/#find-element-from-shadow-root

superchris:
Hey, long pause, but finally getting some cycles to look into this today. Just wanted to check and see if anyone had already done this before I dive in, but doesn't look like it. Let me know if you have any advice on getting me pointed in the right direction

mhanberg:
I have done zero work, so you're in the clear there!

I think the best place to start would be the w3c link i posted above. More or less, there are new endpoints in the webdriver that allow you to query and do stuff with shadow roots.

So, you would add code into the two adapter modules to hit those endpoints, and add functions to the Browser module to use those new functions to do what you want with the shadow roots.

Once you become familiar with it, I would suggest leaving a comment with your proposed API additions, and we can see how they flow with the rest of Wallaby.

I appreciate the effort here, thanks!

mhanberg:
That could be a good place to start.

One idea I just had was to create a `shadow` function on the Query module, and the `Browser.find` function can branch based on the query.

Something like (copying the `find/2` documentation for this example)

```elixir
session
|> find(Query.shadow("#login-button"))
|> assert_text("Login")

# or something like

buttons =
  session
  |> find(Query.css(".login-button", count: 2, text: "Login", shadow: true))
  ```

PR #623, opened 2021-09-02
Fix @spec for `retry`

marcandre:


lib/wallaby/browser.ex
```diff
@@ -141,7 +141,7 @@ defmodule Wallaby.Browser do
   loop forever (or until the test is killed by exunit).
   """
   @type sync_result :: {:ok, any()} | {:error, any()}
-  @spec retry((() -> sync_result), timeout) :: sync_result()
+  @spec retry((() -> sync_result), integer()) :: sync_result()
```
michallepicki, line 144:
```suggestion
  @spec retry((() -> sync_result), non_negative_integer()) :: sync_result()
```

michallepicki (approved):
Hi @marcandre !

I checked the reasoning behind this, for others: `timeout()` is one of [built-in types](https://hexdocs.pm/elixir/typespecs.html#built-in-types) and expands to `:infinity | non_neg_integer()`. This value is used in this private function:

```elixir
  defp max_time_exceeded?(start_time) do
    current_time() - start_time > max_wait_time()
  end
```

where it would not be valid to subtract `:infinity`, so it causes an `ArithmeticError` at runtime.

Thanks for the PR!

lib/wallaby/browser.ex
```diff
@@ -141,7 +141,7 @@ defmodule Wallaby.Browser do
   loop forever (or until the test is killed by exunit).
   """
   @type sync_result :: {:ok, any()} | {:error, any()}
-  @spec retry((() -> sync_result), timeout) :: sync_result()
+  @spec retry((() -> sync_result), non_negative_integer()) :: sync_result()
```
michallepicki, line 144:
```suggestion
  @spec retry((() -> sync_result), non_neg_integer()) :: sync_result()
```

marcandre:
Thanks!

Issue #585, opened 2021-02-10
`has_css?/2` is raising an exception instead of returning `false`


adkron:
## Issue

Description of the issue. Please include browser you're using (phantomjs,
chrome, etc.) elixir and erlang version and any relevant stack traces.

Chome
Wallaby: 0.28.0

## Test Code & HTML

```
refute session
|> visit("/")
|> has_css?("#not_there") 
```

I get a `** (Wallaby.QueryError) Expected to find any, visible elements that matched the css '#apply_discount_153' but 0, visible elements were found.`

`has_no_css?/2` works just fine. It looks like `has_css?/2` uses `count: :any` and `has_no_css?` uses `count: 0`. Also `has_any?/2` uses `find/2` which is where the exception originates.

BrooklinJazz:
Hi @adkron! I haven't contributed to the project before, so forgive me - I'm not up to date on how to submit solutions or proper processes.

I believe I fixed the problem using `has?/1` and added tests for every arity for `has_css?` and `has_no_css?`.

https://github.com/elixir-wallaby/wallaby/pull/624

PR #624, opened 2021-09-09
has_css? should return a boolean. Fix for Issue: #585.

BrooklinJazz:
@adkron posted an issue: https://github.com/elixir-wallaby/wallaby/issues/585 that the `has_css?/2` function was raising an exception.

This is my first PR in the project, and I'm new to open source 🙏 Please let me know if I'm breaking any rules or need to adjust anything!

the `has_css?/2` method was using `find/1` instead of `has?/1` causing it to throw an error instead of returning a boolean.

I've added testing and made test names consistent for has_css?/2, has_css/3, has_no_css?/2, and has_no_css/3 to make sure all work as expected.


michallepicki (approved):
LGTM

PR #633, opened 2021-09-30
Wallaby.Element: handle errors more consistently

jonleighton:
Fixes https://github.com/elixir-wallaby/wallaby/issues/631

A number of functions in Wallaby.Element did not handle {:error, _}
tuples, which meant that they’re raise a CaseClauseError if they
received an error from the driver.

This commit extracts some private functions to make the handling of
driver results more concise and consistent.

I didn’t see any point adding more tests for StaleReferenceError, as
there is already a test for this error, and everything now uses the same
code path.

I removed InvalidSelectorError -- this was only used by
Wallaby.Element.clear/1. As Wallaby.Element is dealing
with elements that we have already found, it’s not possible that we’ll
be handling an invalid selector at this point.

Wallaby.Browser does deal with invalid selectors, but they are raised as
a QueryError, rather than InvalidSelectorError. (There were also no
tests for the latter.)


lib/wallaby/exceptions.ex
```diff
@@ -63,14 +63,6 @@ defmodule Wallaby.StaleReferenceError do
   end
 end
 
-defmodule Wallaby.InvalidSelectorError do
-  defexception [:message]
-
-  def exception(_) do
-    %__MODULE__{message: "Shit is broken and invalid yo"}
```
mhanberg, line 70: 🤨. I didn't know this was in here lol.

lib/wallaby/element.ex
```diff
@@ -284,11 +237,34 @@ defmodule Wallaby.Element do
...
+
+  defp handle_boolean_result(result) do
+    case result do
+      {:ok, true} -> true
+      {:ok, false} -> false
+      {:error, error} -> raise_error(error)
+    end
+  end
+
+  defp raise_error(:stale_reference), do: raise(StaleReferenceError)
```
mhanberg, line 267: WDYT about adding a clause that raises a runtime error with the argument? Might be better than a FunctionClauseError.

mhanberg:
Don't worry about the credo errors. I'll take care of that.

lib/wallaby/element.ex
jonleighton: My thinking was that this would surface errors we're not properly handling, but I'm OK with your suggestion (which would still surface errors we're not properly handling...) :+1:

PR #668, opened 2022-01-25
Attempt to avoid getting confused by these arguments not being %Query{} structs

grzuy:
Hi, thanks for your work on this great library.

Just minor refactor attempt to avoid confusion on what these arguments actually are.

Hope helps.
Thanks.


mhanberg:
Thanks for the contribution.

I don't think selector is a great name to change, since a selector is also another term with another meaning.

Since this is a private module, I am going to leave it as is.

PR #675, opened 2022-03-25
Adds esbuild config info to the README

britton-jb:
Esbuild has become the new standard for Phoenix asset management, and so this adds a note about how to use Wallaby with it.


mhanberg:
Can you add a link to the esbuild hex package?

PR #677, opened 2022-03-25
Adds a note about the on_mount hook for async LiveView testing

britton-jb:
Part 2 for testing LiveView with Wallaby async.
[Part 1](https://github.com/elixir-wallaby/wallaby/pull/676)

### DIFF
--- a/README.md
+++ b/README.md
@@ -243,7 +243,7 @@ This method is less error prone but it will cause a delay when starting your tes
 
 #### LiveView
 
-In order to test Phoenix LiveView with Wallaby you'll also need to add a function to each `mount/3` function in your LiveViews, or use the `on_mount` `live_session` lifecycle hook in the router:
+In order to test Phoenix LiveView (as of [version 0.17.7](https://github.com/phoenixframework/phoenix_live_view/releases/tag/v0.17.7)) with Wallaby you'll also need to add a function to each `mount/3` function in your LiveViews, or use the `on_mount` `live_session` lifecycle hook in the router:
 
 ```elixir
 defmodule MyApp.Hooks.AllowEctoSandbox do
--- a/README.md
+++ b/README.md
@@ -241,6 +241,38 @@ The second solution is to add a new alias to your mix config that recompiles ass
 
 This method is less error prone but it will cause a delay when starting your test suite.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -214,6 +214,38 @@ The second solution is to add a new alias to your mix config that recompiles ass
...
+    Phoenix.Ecto.SQL.Sandbox.allow(metadata, Ecto.Adapters.SQL.Sandbox)
+  end
+end
+```
+
+and then including the function usage in the router:
+
+```elixir
+live_session :default, on_mount: MyApp.Hooks.AllowEctoSandbox do
+  ...
```
mhanberg, line 245:
```suggestion
  # ...
```

mhanberg:
One last thought, sorry. 

What do you think about adding the current live view version to the docs you added? 

Just so if it changes and becomes out of date, it will hopefully be fairly obvious to check the love view changelog.

britton-jb:
That’s a great call, I’ll add that. Thanks.

README.md
```diff
@@ -214,6 +214,38 @@ The second solution is to add a new alias to your mix config that recompiles ass
...
+    {:cont, socket}
+  end
+
+  defp allow_ecto_sandbox(socket) do
+    %{assigns: %{phoenix_ecto_sandbox: metadata}} =
+      assign_new(socket, :phoenix_ecto_sandbox, fn ->
+        if connected?(socket), do: get_connect_info(socket, :user_agent)
+      end)
+
+    Phoenix.Ecto.SQL.Sandbox.allow(metadata, Ecto.Adapters.SQL.Sandbox)
```
jonleighton, line 236: Nice one for doing this! I had to set it up last week and having this explicit documentation will really help.

One thing to add is that we currently recommend using a config key to define the sandbox:

``` elixir
config :your_app, :sandbox, Ecto.Adapters.SQL.Sandbox
```

So then this line could be:

``` elixir
Phoenix.Ecto.SQL.Sandbox.allow(metadata, Application.get_env(:your_app, :sandbox))
```

That then makes it continue to work properly if/when you define a custom sandbox module to be able to sandbox other things than the database connection (e.g. Mox mocks -- see the current Wallaby docs).

PR #695, opened 2022-06-24
Add sass asset note to README

eToThePiIPower:
The **Assets/esbuild** section of the README includes a step on adding `esbuild default` to the `test` alias. This works to recompile javascript files on every run; but it only compiles scss styles on the *first* run, leading to continued confusion when debugging.

To do this, we simply need to add the `sass default` step to the `test` alias.

It's wasn't terribly difficult to figure out, but could save other people the 15 minutes and seems to fit well into this section.


mhanberg:
This would probably be more helpful if the information is agnostic to the asset tool that is being used.

Would you mind rephrasing the section to not directly mention esbuild or sass?

PR #698, opened 2022-07-15
fix(chromedriver): Account for Chromium when doing the version matching

### LINKED ISSUE #697: Allow to skip chromedriver-chrome version match check
After https://github.com/elixir-wallaby/wallaby/pull/688 got merged, Wallaby no longer works with chromium on Linux (I think it will not work with a custom installation path on Windows either).

One solution would be to allow the chromedirver-chrome version match check to be skipped.

mhanberg:
We weren't accounting for Chromium when checking the versions of Chrome and ChromeDriver were matching.

@michallepicki can you install this branch and see if it fixes the problem? I was able to confirm that it works with Chromium locally on my MacBook, but haven't tested it manually for Chromium on Linux.

Thanks

Fixes #697 (maybe)

### DIFF
--- a/lib/wallaby/chrome.ex
+++ b/lib/wallaby/chrome.ex
@@ -174,7 +174,7 @@ defmodule Wallaby.Chrome do
             "HKLM\\SOFTWARE\\Wow6432Node\\Microsoft\\Windows\\CurrentVersion\\Uninstall\\Google Chrome"
           ])
 
-        chrome_version = parse_version("Version", stdout)
+        chrome_version = parse_version(stdout)
 
         {:ok, chrome_version}
 
@@ -182,7 +182,8 @@ defmodule Wallaby.Chrome do
         case find_chrome_executable() do
           {:ok, chrome_executable} ->
             {stdout, 0} = System.cmd(chrome_executable, ["--version"])
-            chrome_version = parse_version("Google Chrome", stdout)
+
+            chrome_version = parse_version(stdout)
 
             {:ok, chrome_version}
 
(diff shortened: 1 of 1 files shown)


conceivably:
As far as I can tell, I had the same problem with Chromium on Linux on the Wallaby main branch, in that I got:

```
** (MatchError) no match of right hand side value: {:error, {:wallaby, {:bad_return, {{Wallaby, :start, [:normal, []]}, {:EXIT, {%Wallaby.DependencyError{message: "Wallaby can't find Chrome. Make sure you have chrome installed and included in your path.\nYou can also provide a path using `config :wallaby, :chromedriver, binary: <path>`.\n"}, [{Wallaby, :start, 2, [file: 'lib/wallaby.ex', line: 33, error_info: %{module: Exception}]}, {:application_master, :start_it_old, 4, [file: 'application_master.erl', line: 293]}]}}}}}}
    test/test_helper.exs:4: (file)
    (elixir 1.13.1) lib/code.ex:1183: Code.require_file/2
    (elixir 1.13.1) lib/enum.ex:937: Enum."-each/2-lists^foreach/1-0-"/2
```

With this branch the problem disappears for me :+1:
Thank you for the fix :)

mhanberg:
@conceivably im not sure that it matters but are you running this locally or just CI?

johnf9896:
I would like to add that in Alpine Linux the Chromium binary is called `chromium-browser` instead of `chromium`. This branch fixes it for me locally, but not for Docker (used both locally and on CI).

mhanberg:
@johnf9896 would you be able to provide me with the docker file that demonstrates the failure?

johnf9896:
I guess I could create a simple project for it but I think the problem can be solved by adding `"chromium-browser" to `["google-chrome", "chromium"]`.  Everything works if I specify the path manually, otherwise I get the same error as @conceivably.

mhanberg:
@johnf9896 Turns out that I had just missed that name when I was looking through the chromedriver codebase for the list of binary locations that it will look for. `chromium-browser` was there.

I pushed up another commit that adds that name.

johnf9896:
That indeed solves it :)

mhanberg:
Thanks for all the help folks.

This seems like it's fixed, but I'll wait for @michallepicki to confirm.

michallepicki:
this branch starts Chromium for me as well, thanks

PR #711, opened 2022-10-01
FEATURE: Allow setting of optional cookie attributes

merhard:
This PR adds a `set_cookie/4` function to permit the setting of optional cookie attributes per the [WebDriver specification](https://www.w3.org/TR/webdriver1/#cookies).

I have a project that needs to be able to set more than just `name` and `value` of a cookie; namely `Domain`, `Secure`, and `HttpOnly`.


mhanberg:
This is great, thanks!

PR #713, opened 2022-11-09
Update `send_keys/2` docs to use correct module

vanderhoop:
I copy-pasted an example from the `Wallaby.Browser` hexdocs page, only to find that the example for `send_keys/2` was using the wrong module namespace.


mhanberg:
💛🦘💜

PR #714, opened 2022-11-09
Surface 'text' condition in css query error message when specified

vanderhoop:
## Summary
This pull request surfaces the `text` condition value when it's specified in CSS queries, and leaves the existing error message as-is when the condition _isn't_ specified.

## Motivation

I frequently build out css queries using the `text` condition (in combination with the base css selector of the query). Sometimes I would get failures where the element selector was found in the application UI but failed due the `text` condition, but in those cases the error message would only reference the selector, which would lead me to think that the element hadn't been present at all. 

These cases usually led me to open the screenshots and see the element present in the UI and realize my error—usually a case mismatch or outright typo🤦🏻‍♂️—and it's happened to me enough that I figured the lib could and should specify this condition in the error.

When the text-condition _isn't_ specified, we leave the existing error message as-is so as to not confuse devs.

## Before

image

## After

image

### DIFF
--- a/lib/wallaby/query/error_message.ex
+++ b/lib/wallaby/query/error_message.ex
@@ -118,9 +118,13 @@ defmodule Wallaby.Query.ErrorMessage do
     "'#{name}' with value '#{value}'"
   end
 
-  def selector(%Query{selector: selector}) do
-    "'#{selector}'"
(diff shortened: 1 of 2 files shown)


mhanberg:
Excellent!

cysiegel:
Thank you - this feature saves a lot of time when debugging

Issue #720, opened 2022-12-31
One chromedriver per core becomes problematic with sufficiently many cores


chriscrabtree:
### Elixir and Erlang/OTP versions

Erlang/OTP 25 [erts-13.1.2] [source] [64-bit] [smp:56:56] [ds:56:56:10] [async-threads:1] [jit:ns]

Elixir 1.14.2 (compiled with Erlang/OTP 25)

### Operating system

macOS 12.6

### Browser

n/a

### Driver

ChromeDriver

### Correct Configuration

- [X] I confirm that I have Wallaby configured correctly.

### Current behavior

To reproduce, run `mix test` on Wallaby itself.

On my 10-core MacBook Pro, tests run quickly (~11 seconds)  and everything passes.

On my 56-core Mac Pro desktop, tests run very slowly (~46 seconds) with 7 failures and many timeout messages.

[out.txt](https://github.com/elixir-wallaby/wallaby/files/10327662/out.txt)

### Expected behavior

The Wallaby test suite should pass quietly on the 56-core Mac Pro.

### Test Code & HTML

I fixed the issue locally by changing the `partitions` value in the `init/1` function of `lib/wallaby/chrome.ex`.

Rather than `partitions: System.schedulers_online()`, I have it as:

`partitions: reasonable_number_of_partitions(System.schedulers_online())`

And then:
```
  defp reasonable_number_of_partitions(schedulers_online, divisor \\ 6)

  defp reasonable_number_of_partitions(schedulers_online, divisor)
       when is_integer(schedulers_online) and schedulers_online >= 56 and is_integer(divisor) do
    Integer.floor_div(schedulers_online, divisor)
  end

  defp reasonable_number_of_partitions(schedulers_online, _divisor)
       when is_integer(schedulers_online) do
    schedulers_online
  end
```

This maintains the current Wallaby behavior of using all cores unless there are at least 56 cores, in which case I experimentally estimated that one-sixth of the cores is the optimum for performance, at least on the Wallaby test suite.

I suspect that the `56` core threshold should actually be lower, but I don't have any machines between 10-core and 56-core to test that.

### Configuration?
If we made some of these options configurable, then users could have an escape hatch if they were to run into a similar situation in the future.

Happy to help with a PR if we can land on a good approach for that configuration. I'm not so familiar with Wallaby internals that I know how we would want to make it more configurable in that regard.

### A (Very) Incremental Improvement
[...]

PR #722, opened 2023-01-08
Add partition count config and improved defaults (#720)

chriscrabtree:
The user can explicitly set the number of chromedriver partitions now.

But even if the user does not set this config, Wallaby will make an improved choice as to how many partitions it fires up.

This will lead to better performance and fewer false negative test results on machines with high CPU-core counts.


mhanberg (changes requested):
...

I left some comments on this PR weeks ago, but never clicked "Submit review"...

Sorry about that.

chriscrabtree:
Understood, and done. I did find that I had to implement the same `min(System.schedulers_online(), 10)` idea in a few different places so that running `mix test` on Wallaby itself on a high-core-count machine would run smoothly.

While I felt a strong urge to factor that out into a meaningfully named function, lol, I went with the direct approach in accordance with the earlier guidance.

mhanberg:
@chriscrabtree I pushed up some changes, can you confirm that it works alright with your computer with tons of cores, and I will merge and release this.

chriscrabtree:
@mhanberg Hey, I appreciate you thinking about the possible consequences--good catch!

Indeed, it seems that at least one of the three configurations removed are in fact necessary for a smooth test run on the 56-core machine.

This is the [recording of the run after I merged the latest commits](https://asciinema.org/a/Gf42Weyzwa4Pu6SrIYgQn1YZU). What this recording doesn't show is spiking all the cores while it slowly grinds through the test suite with multiple failures.

And [this is the recording after I reapplied the essential 10-max logic](https://asciinema.org/a/Lw7aVYVvrAhEZZ9Y33Ho5S6Ub) in those three same places, which is both 100% successful and ~4x faster.

I can play with identifying the critical set of changes only to allow it to run smoothly.

And just so I'm clear: we no longer want to have concurrency as a configurable option?

chriscrabtree:
@mhanberg Okay, I determined that adding the 10-max governor on PartitionSupervisor is the one truly necessary change. Just pushed a commit reflecting that single thing which allows a smooth mix test experience.

It's possible that as the test suite grows over time, we might need to adjust ExUnit :max_cases and the :wallaby_pool size as well, but after multiple runs with the current suite, I noted no failures. So we can cross that bridge if and when we ever come to it.

chriscrabtree:
Realizing now those asciinema recordings may have been private by default. I've now set them public.

mhanberg:
@chriscrabtree I had missed something when i pushed up my original patches when simplifying the solution. You can see the commit message for more details.

Can you pull this down again and just re-confirm that it still works? (It should)

chriscrabtree:
@mhanberg Confirmed!

mhanberg:
This has been released with [v0.30.3](https://diff.hex.pm/diff/wallaby/0.30.2..0.30.3)

PR #728, opened 2023-03-15
Shadow dom

superchris:
@mhanberg this is what we ended up with. We added a `shadow_root/1` function on `Wallaby.Browser` that takes an element and narrows the scope to the shadow root of the element. This seems to line up really nicely with what webdriver wants and pretty much lets everything "just work" in what seems like a nice way IMO. Curious to hear your thoughts. If you think it's ok I'll add docs and get it mergable

### DIFF
--- a/integration_test/cases/browser/shadow_dom_test.exs
+++ b/integration_test/cases/browser/shadow_dom_test.exs
@@ -13,10 +13,11 @@ defmodule Wallaby.Integration.Browser.ShadowDomTest do
   end
 
   test "can find a shadow root", %{session: session} do
-    element =
+    shadow_root =
       session
-      |> find_shadow(Query.css("shadow-test"))
+      |> find(Query.css("shadow-test"))
+      |> shadow_root()
 
-    assert element
+    assert shadow_root
   end
 end
--- a/integration_test/cases/browser/shadow_dom_test.exs
+++ b/integration_test/cases/browser/shadow_dom_test.exs
@@ -20,4 +20,21 @@ defmodule Wallaby.Integration.Browser.ShadowDomTest do
 
     assert shadow_root
   end
+
+  test "can find stuff within da shadow dom", %{session: session} do
+    element =
+      session
+      |> find(Query.css("shadow-test"))
+      |> shadow_root()
+      |> find(Query.css("#in-shadow"))
+  end
+
+  test "can click stuff within da shadow dom", %{session: session} do
+    element =
+      session
+      |> find(Query.css("shadow-test"))
+      |> shadow_root()
+      |> click(Query.css("button"))
+  end
+
 end
--- a/integration_test/cases/browser/shadow_dom_test.exs
+++ b/integration_test/cases/browser/shadow_dom_test.exs
@@ -1,0 +1,22 @@
+defmodule Wallaby.Integration.Browser.ShadowDomTest do
+
+  use Wallaby.Integration.SessionCase, async: true
+
+  import Wallaby.Query, only: [css: 1]
+
+  setup %{session: session} do
+    page =
+      session
+      |> visit("shadow_dom.html")
+
+    {:ok, page: page}
+  end
+
+  test "can find a shadow root", %{session: session} do
+    element =
+      session
+      |> find_shadow(Query.css("shadow-test"))
+
+    assert element
+  end
+end
--- a/lib/wallaby/browser.ex
+++ b/lib/wallaby/browser.ex
@@ -1013,7 +1013,6 @@ defmodule Wallaby.Browser do
   end
 
   def shadow_root(%{driver: driver} = element) do
-    IO.inspect(driver)
     driver.shadow_root(element)
   end
 
--- a/lib/wallaby/browser.ex
+++ b/lib/wallaby/browser.ex
@@ -999,6 +999,26 @@ defmodule Wallaby.Browser do
     parent
   end
 
+  def find_shadow(%{driver: driver}, query) do
+    retry(fn ->
+      try do
+        with {:ok, query} <- Query.validate(query),
+             compiled_query <- Query.compile(query),
+             {:ok, elements} <- driver.find_shadow(parent, compiled_query),
+             {:ok, elements} <- validate_visibility(query, elements),
+             {:ok, elements} <- validate_text(query, elements),
+             {:ok, elements} <- validate_selected(query, elements),
+             {:ok, elements} <- validate_count(query, elements),
+             {:ok, elements} <- do_at(query, elements) do
+          elements
+        end
+      rescue
+        StaleReferenceError ->
+          {:error, :stale_reference}
+      end
+    end)
+  end
+
   @doc """
   Finds all of the DOM elements that match the CSS selector. If no elements are
   found then an empty list is immediately returned. This is equivalent to calling
--- a/lib/wallaby/browser.ex
+++ b/lib/wallaby/browser.ex
@@ -999,26 +999,6 @@ defmodule Wallaby.Browser do
     parent
   end
 
-  def find_shadow(%{driver: driver}, query) do
-    retry(fn ->
-      try do
-        with {:ok, query} <- Query.validate(query),
-             compiled_query <- Query.compile(query),
-             {:ok, elements} <- driver.find_shadow(parent, compiled_query),
-             {:ok, elements} <- validate_visibility(query, elements),
-             {:ok, elements} <- validate_text(query, elements),
-             {:ok, elements} <- validate_selected(query, elements),
-             {:ok, elements} <- validate_count(query, elements),
-             {:ok, elements} <- do_at(query, elements) do
-          elements
-        end
-      rescue
-        StaleReferenceError ->
-          {:error, :stale_reference}
-      end
-    end)
-  end
-
   @doc """
   Finds all of the DOM elements that match the CSS selector. If no elements are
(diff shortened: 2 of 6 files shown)


mhanberg:
for posterity and myself, would you mind commenting with a lay person's explanation of what the shadow root is and how wallaby has to talk with the webdriver in order for the user to interact with them?

integration_test/cases/browser/shadow_dom_test.exs
```diff
@@ -0,0 +1,48 @@
+defmodule Wallaby.Integration.Browser.ShadowDomTest do
```
mhanberg, line 1: I believe you need to add this test to `integration_tests/tests.exs` for them to actually run.

superchris: Thanks!

superchris:
> for posterity and myself, would you mind commenting with a lay person's explanation of what the shadow root is and how wallaby has to talk with the webdriver in order for the user to interact with them?

Sure! Here is my attempt: Shadow DOM is a sub-specification of the Web Components group of specs that allow a developer to create an encapsulated DOM within an element. This encapsulated DOM, or shadow DOM, can contain CSS, markup, and anything else a DOM can contain but is isolated from the rest of the document such that styles from the containing DOM do not apply to the shadow DOM (with specific exceptions) and vice versa. This allows a developer of a custom element to have complete control over the precise rendering of their element while avoiding collisions with or dependencies on the containing document's CSS.

Because the Shadow DOM is isolated from the containing DOM, queries within the containing document will not find elements in the shadow DOM of a custom element. In order to overcome this limitation, several new API calls have been added to the WebDriver protocol to "pierce" a shadow DOM and query elements within it. The way we interact with these in wallaby is to scope our query to a shadow DOM of a specific element by using the `shadow_root` function, eg:

```elixir
    element =
      session
      |> find(Query.css("shadow-test"))
      |> shadow_root()
      |> find(Query.css("#in-shadow"))
```

This will query for an element with id `in-shadow` that is within the shadow DOM for a `<shadow-test>` element. The name `shadow_root` corresponds the DOM property `shadowRoot` which is the accessor property for a shadow DOM within an element.

mhanberg:
Excellent explanation!!

I have a question about the API. 

When we query an element that contains shadow dom, is it possible to know that before we call the shadow_dom function?

If we can mark an element has containing shadow dom, we can potentially hide the shadow_dom function from the user and they can just pretend that they are querying inside of it like normal.

superchris:
Thanks! I don't know of a way to detect if an element has a shadow root, other than with javascript or doing the api call I am doing already in `shadow_root`. Doing this implicitly for every time we return or scope within an element seems tricky to me. I would be more in favor of being explicit I think. I also think the use case is fairly specific, my guess is for now the percentage of people using wallaby to interact with shadow DOM is small so adding another api call all the time seems maybe not the best tradeoff.

superchris:
@mhanberg in doing some further experimentation, I came across an example where we really need to know whether the user is intending to query inside the shadow dom or the 'light dom' of a custom elements. Custom elements can also have markup placed inside them in the surrounding document, but then have this content merged into the shadow DOM using a feature called slots. This gets complicated, but the result is the element has both light and shadow DOMs :) To query for things in this "light dom" you do *not* want to scope using `shadow_root`. I'll try to come up with a test case that demonstrates this.

integration_test/cases/browser/shadow_dom_test.exs
```diff
@@ -0,0 +1,49 @@
...
+    {:ok, page: page}
+  end
+
+  test "can find a shadow root", %{session: session} do
+    shadow_root =
+      session
+      |> find(Query.css("shadow-test"))
+      |> shadow_root()
+
+    assert shadow_root
```
mhanberg, line 18: Will this be an Element struct? if yes let's assert on that

integration_test/cases/browser/shadow_dom_test.exs
```diff
@@ -0,0 +1,49 @@
...
+  end
+
+  test "can find stuff within da shadow dom", %{session: session} do
```
mhanberg, line 21: let's make these test names slightly more formal

integration_test/cases/browser/shadow_dom_test.exs
```diff
@@ -0,0 +1,49 @@
...
+    element =
+      session
+      |> find(Query.css("shadow-test"))
+      |> shadow_root()
+      |> find(Query.css("#in-shadow"))
+
+    assert element
```
mhanberg, line 28: let's actually assert on the contents of these values rather than just on truthiness

lib/wallaby/browser.ex
```diff
@@ -1012,6 +1012,26 @@ defmodule Wallaby.Browser do
     )
   end
 
+  @doc """
+  Finds the shadow DOM for the specified element and returns an element corresponding
+  to this shadow DOM. Subsequent queries made within element retured by calling `shadow_root`
+  will return elements within the shadow DOM, e. g.
```
mhanberg, line 1018:
```suggestion
  Finds and returns the shadow root for the given element.
  
  Queries executed on the returned shadow root will be scoped to the root's shadow DOM.
```

lib/wallaby/browser.ex
```diff
@@ -1012,6 +1012,26 @@ defmodule Wallaby.Browser do
...
+
+  ```
+  element =
+    session
+    |> find(Query.css("shadow-test"))
+    |> shadow_root()
+    |> find(Query.css("#in-shadow"))
+  ```
```
mhanberg, line 1026:
```suggestion
  ```elixir
  session
  |> find(Query.css("shadow-test"))
  |> shadow_root()
  |> find(Query.css("#in-shadow"))
  ```
```

lib/wallaby/browser.ex
```diff
@@ -1012,6 +1012,26 @@ defmodule Wallaby.Browser do
...
+  """
+  def shadow_root(%{driver: driver} = element) do
```
mhanberg, line 1028: please add an `@spec` annotation

lib/wallaby/driver.ex
```diff
@@ -180,6 +180,11 @@ defmodule Wallaby.Driver do
   @callback find_elements(Session.t() | Element.t(), Query.compiled()) ::
               {:ok, [Element.t()]} | {:error, reason}
 
+  @doc """
+  Invoked to find shadow root of given element
```
mhanberg, line 184:
```suggestion
  Invoked to find shadow root of the given element.
```

lib/wallaby/webdriver_client.ex
```diff
@@ -48,6 +49,18 @@ defmodule Wallaby.WebdriverClient do
          do: {:ok, elements}
   end
 
+  @doc """
+  Finds the shadow root for a given element.
```
mhanberg, line 53:
```suggestion
  Finds the shadow root for the given element.
```

lib/wallaby/webdriver_client.ex
```diff
@@ -48,6 +49,18 @@ defmodule Wallaby.WebdriverClient do
...
+  """
+  @spec shadow_root(Element.t()) :: {:ok, Element.t()}
+  def shadow_root(element) do
+    with {:ok, resp} <- request(:get, element.url <> "/shadow"),
+         {:ok, shadow_root} <- Map.fetch(resp, "value"),
+         element <- cast_as_element(element, shadow_root) do
+      {:ok, element}
+    end
+  end
```
mhanberg, line 62:
```suggestion
  def shadow_root(element) do
    with {:ok, resp} <- request(:get, element.url <> "/shadow"),
         {:ok, shadow_root} <- Map.fetch(resp, "value") do
      {:ok, cast_as_element(element, shadow_root)}
    end
  end
```

test/wallaby/http_client_test.exs
```diff
@@ -55,6 +55,22 @@ defmodule Wallaby.HTTPClientTest do
       assert {:error, :stale_reference} = Client.request(:post, bypass_url(bypass, "/my_url"))
     end
 
+    test "with a non-existant shadow root response", %{bypass: bypass} do
```
mhanberg, line 58:
```suggestion
    test "with a non-existent shadow root", %{bypass: bypass} do
```

mhanberg, line 58: please add a success case test

lib/wallaby/chrome.ex
```diff
@@ -410,6 +410,9 @@ defmodule Wallaby.Chrome do
   defdelegate accept_prompt(session, input, fun), to: WebdriverClient
   @doc false
   defdelegate dismiss_prompt(session, fun), to: WebdriverClient
+  @doc false
+  defdelegate shadow_root(element), to: WebdriverClient
+
```
mhanberg, line 415:
```suggestion
```

mhanberg (changes requested):
This is looking good.

I think for another PR, it would be good start create a guide that gives a more in depth example. You can see an example of how to add a guide [here](https://github.com/mhanberg/temple/blob/7a505875af6a1cee1536e516528f5be914df1f3f/mix.exs#L41-L55)

Timmitry:
Hello 👋 What is the state of this PR, is this still something you would be open to add support for? What is still missing to get this merged?

dklotz:
Hi @mhanberg - did the last commit by @lessthanseventy address all your comments? We are using WebComponents in a project and could really benefit from a Wallaby release that included the `|> shadow_root()` functionality.

PR #730, opened 2023-04-17
execute_cdp_command added

hissssst:
Please help me write the test.

I was able to reproduce that this change worked correctly with
```
echo "HTTP/1.1 200 OK\r\nServer: netcat\r\n\r\n" | nc -l 9999
```
And then
```elixir
iex> {:ok, session} = Wallaby.start_session()
iex> Wallaby.Chrome.execute_cdp_command(session, "Network.setUserAgentOverride", %{"userAgent" => "hey"})
iex> Wallaby.visit(session, "http://localhost:9999/")
```

And then netcat returned
```
GET /x HTTP/1.1
Host: localhost:9999
Connection: keep-alive
Upgrade-Insecure-Requests: 1
User-Agent: hey
Accept: text/html,application/xhtml+xml,application/xml;q=0.9,image/avif,image/webp,image/apng,*/*;q=0.8,application/signed-exchange;v=b3;q=0.7
Sec-Fetch-Site: none
Sec-Fetch-Mode: navigate
Sec-Fetch-User: ?1
Sec-Fetch-Dest: document
Accept-Encoding: gzip, deflate, br
Accept-Language: en-US
```

### DIFF
--- a/lib/wallaby/chrome.ex
+++ b/lib/wallaby/chrome.ex
@@ -527,6 +527,29 @@ defmodule Wallaby.Chrome do
     end
   end
 
+  @doc """
+  Chromium specific function to execute CDP (Chrome Devtools Protocol) commands
+  https://chromedevtools.github.io/devtools-protocol/
+  """
+  @spec execute_cdp_command(Wallaby.Session.t(), String.t(), map()) :: {:ok, any()} | {:error, any()}
+  def execute_cdp_command(session, command, params) do
+    url = "#{session.session_url}/goog/cdp/execute"
+    case Wallaby.HTTPClient.request(:post, url, %{cmd: command, params: params}) do
+      {:ok, %{"status" => 0, "value" => value}} ->
+        {:ok, value}
+
+      {:ok, %{"status" => status} = response} ->
+        {:error, {:bad_status, status, response}}
+
+      {:ok, %{}} ->
+        # Value not presend
+        :error
+
+      other ->
+        other
+    end
+  end
+
   @doc false
   def find_elements(session_or_element, compiled_query),
     do: delegate(:find_elements, session_or_element, [compiled_query])


mhanberg:
Can you link to any documentation of the feature you are trying to implement?

I cursor search leads me to believe that this is a Selenium specific feature, but I'm not entirely sure.

hissssst:
@mhanberg  https://chromedevtools.github.io/devtools-protocol/
It's in the PR

mhanberg:
I am aware of what the Chrome Devtools protocol is, but I am asking for documentation on how you use Chromedriver or Selenium with it.

For example, your patch makes a request to a `/goog/cdp/execute` path, but I have no idea where that came from or where it is documented.

hissssst:
I've just took a look at Selenium implementation

* Here is the call format
https://github.com/SeleniumHQ/selenium/blob/678854bfa08dcc0bee87a94a22f771e56fc72305/py/selenium/webdriver/chromium/webdriver.py#L169
* Here's the path
https://github.com/SeleniumHQ/selenium/blob/678854bfa08dcc0bee87a94a22f771e56fc72305/py/selenium/webdriver/chromium/remote_connection.py#L38
* And the `goog` is a `vendor_prefix` for chromedriver

mhanberg:
Where is the actual documentation though?

Also, it seems like you added code to the Chromedriver driver, but based on what you've shared so far, it seems like a Selenium specific feature. Selenium will use Chromedriver, but you wouldn't add the code to the Wallaby's Chromedriver driver, as it uses Chromedriver directly

hissssst:
Can you help me with this please, what kind of documentation do you expect?
I am unfamiliar with chromedriver and selenium relations, so I can't really answer your question without any clarification

mhanberg:
@hissssst if I knew where the documentation was, I wouldn't be requesting it haha.

I am expecting any kind of web page that details this endpoint. I can't be expected to approve or merge code for a feature that I know nothing about and can't find anywhere on the internet.

hissssst:
I think that since this endpoint is not a part of w3's webdriver protocol and there's no description of any endpoint in selenium, there is no documentation except implementations above :\

Issue #569, opened 2020-09-01
Ensure selenium image v4 works


peaceful-james:
As per https://elixirforum.com/t/setting-up-wallaby-with-docker-compose-chrome-and-selenium/32296/14?u=slouchpie

i am opening a reminder issue to test against v4 selenium.

michallepicki:
Related: https://github.com/elixir-wallaby/wallaby/issues/455
I think Selenium v4 might have deprecated the JSON Wire Protocol and now requires the W3C one, but that will have to be confirmed.

blackham:
Not sure if this is a Docker HUB v4 issue or Selenium itself (or my bad programming). It looks like it's a wallaby issue since it works with Ruby. 

Here is what worked and didn't: 
single selenium v4 docker instance with Ruby "selenium-webdriver"  - Good
single selenium v4 docker instance with Wallaby  - Good
selenium-hub v4 docker-compose with Ruby "selenium-webdriver" - Good
selenium-hub v4 docker-compose with Wallaby - Sub-par results

### Start TL;DR;
Wallaby results with v4 hub:
Wallaby.start_session returns :ok but the commands I send to that session fail. The error from the selenium-hub log includes this bit: 

`..."attributes": {"exception.message": "Unable to execute request for an existing session: Unable to find session with ID: \nBuild info: version: '4.0.0-rc-2',...`

My guess is the ID isn't being loaded, stored or sent correctly by Wallaby. 
### End TL;DR;

To reproduce: From LInux install docker and docker-composer (pacman -S docker docker-composer). Configure docker to use TCP (see: https://gist.github.com/styblope/dc55e0ad2a9848f2cc3307d4819d819f .. if any selenium docker dev guys read this, PLEASE add this note to your README.md) 

Create a hub config file in /tmp/config.toml

```
[docker]                                                                                                                                                                         
configs = [
    "selenium/standalone-firefox:4.0.0-rc-2-prerelease-20210923", "{\"browserName\": \"firefox\"}",
    "selenium/standalone-chrome:4.0.0-rc-2-prerelease-20210923", "{\"browserName\": \"chrome\"}"
    ]

url = "http://172.17.0.1:2375"
video-image = "selenium/video:ffmpeg-4.3.1-20210923"
```

Create a place to save the sessions (video?):
mkdir /tmp/assets


Create a docker-compose file /tmp/docker-compose-v3-dynamic-grid.yml
```
version: "3"
services:
  node-docker:
[...]

PR #738, opened 2023-07-13
Update Dependencies and test on Elixir 1.14 and 1.15

kyleboe:


kyleboe:
I opened up a [PR](https://github.com/kyleboe/wallaby/pull/1) on my fork so I could run CI and it seems to be failing for the same reason as the Selenium v4 macOS pipeline. It is pulling a newer version of elixir and there are deps that are incompatible with elixir 1.14+

mhanberg:
@kyleboe the v4 jobs aren't expected to succeed. I just have them in there in case someone wants to look into #569

.github/workflows/ci.yml
```diff
@@ -10,7 +10,11 @@ jobs:
 
     strategy:
       matrix:
-        elixir-otp: [{otp: 24.x, elixir: 1.12.x}, {otp: 25.x, elixir: 1.13.x}]
+        elixir-otp:
+          - otp: 25.x
+            elixir: 1.14.x
+          - otp: 26.x
+            elixir: 1.15.x
```
mhanberg, line 17: why did you remove 1.12 and 1.13?

test/wallaby/http_client_test.exs
```diff
@@ -73,7 +73,7 @@ defmodule Wallaby.HTTPClientTest do
 
     test "includes the original HTTPoison error when there is one", %{bypass: bypass} do
       expected_message =
-        "Wallaby had an internal issue with HTTPoison:\n%HTTPoison.Error{id: nil, reason: :econnrefused}"
+        "Wallaby had an internal issue with HTTPoison:\n%HTTPoison.Error{reason: :econnrefused, id: nil}"
```
mhanberg, line 76: if this is flaky, i would do multiple `=~` asserts instead

.github/workflows/ci.yml
kyleboe: Because they are not backwards compatible with the changed dependencies.

kyleboe: This will likely require a major version bump to signify this change.

test/wallaby/http_client_test.exs
kyleboe: I'm not confident that it is flaky just yet. I want to understand why this change was necessary.

mhanberg: otp changed how maps work for keys < 32 so now key order is not deterministic like it used to be

.github/workflows/ci.yml
mhanberg: I don't want to drop support for older versions of Elixir just for a dev dependency that I have never even used since taking over the project.

you can probably add `--only test` to mix deps.get in CI to avoid having to update benchee.

test/wallaby/http_client_test.exs
kyleboe: Agreed. But if maps < 32 keys are treated like keyword lists in memory, they are linked and therefore have a deterministic order yes? So my inference would be that `{:id, nil}` was added before `{:reason, :econnrefused}` in the internals of HTTPoison, thereby putting `:reason` at the head of the linked list.

.github/workflows/ci.yml
kyleboe: I am down to drop benchee as a dependency altogether. However there are other changes around `ExUnit.Server.modules_loaded()` that are not backwards compatible.

test/wallaby/http_client_test.exs
mhanberg: I'm trying to say that in OTP26 that is no longer the case. the order is nondeterministic

https://www.erlang.org/blog/otp-26-highlights/#changed-ordering-of-atom-keys

You can pass `custom_options: [sort_maps: true]` to as an Inspect.Opt tho, which will sort it and it should be deterministic.

.github/workflows/ci.yml
mhanberg: that is a private elixir API, and is in a unit test.

I would just add a conditional that checks if the function is exported and call the one that is

`function_exported?(ExUnit.Server, :modules_loaded, 0)`, etc

test/wallaby/http_client_test.exs
kyleboe: Ah. I see! Thanks for explaining 😃 I'll make some more adjustments and see if I can get things backwards compatible as well. Thanks for the feedback!

mhanberg: Of course 😁

PR #758, opened 2024-01-10
refactor to map_intersperse

bradhanks:


config/dev.exs
```diff
@@ -1 +1 @@
-use Mix.Config
```
mhanberg, line 1: Can you revert these config changes

bradhanks: yes i'm submitting PR now.

mhanberg:
thanks!

PR #759, opened 2024-01-10
Fix Wallaby.Element.size/1 spec

NikitaNaumenko:
Hi, found a minor bug, spec for function size is wrong.

Thanks!


mhanberg:
Thanks!

Issue #761, opened 2024-03-26
Suggested Phoenix Configuration for Concurrent Feature Test Issues Warning


stratigos:
### Elixir and Erlang/OTP versions

Erlang/OTP 26 [erts-14.2.2] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:1] [jit:ns]

Elixir 1.16.1 (compiled with Erlang/OTP 24)

### Operating system

Ubuntu 22.04

### Browser

N/A

### Driver

chromedriver 123

### Correct Configuration

- [X] I confirm that I have Wallaby configured correctly.

### Current behavior

When following the current installation guide from here: https://github.com/elixir-wallaby/wallaby?tab=readme-ov-file#phoenix, specifically this section:

```elixir
  if sandbox = Application.get_env(:your_app, :sandbox) do
    plug Phoenix.Ecto.SQL.Sandbox, sandbox: sandbox
  end
```

The following warning is issued:

```sh
    warning: Application.get_env/2 is discouraged in the module body, use Application.compile_env/3 instead
    │
 10 │   if sandbox = Application.get_env(:my_app, :sandbox) do
    │                            ~
    │
    └─ lib/myapp_web/endpoint.ex:10:28: MyAppWeb.Endpoint (module)
```

### Expected behavior

The documentation has up-to-date instructions that do not result in a warning.

### Test Code & HTML

No test needed, simply run app after following `wallaby` README for installing with a Phoenix app

ca. https://github.com/elixir-wallaby/wallaby/commit/cab3b1fbd8f0bcbcd72ac49f33f339ed71a593e5

mhanberg:
PR welcome

PR #762, opened 2024-03-26
Update README to Avoid Elixir Warning

stratigos:
A warning is issued for use of `get_env/2` used outside of a module. The warning suggests using `compile_env/3` instead.

This change updates the documentation to conform to this suggestion and remove the warning, resulting in a cleaner and more modern setup for `wallaby`.

Resolves [#761](https://github.com/elixir-wallaby/wallaby/issues/761)


stratigos:
This worked just fine in my tests and when running my Phoenix application.

PR #765, opened 2024-04-04
Update README: Test Sandbox File Location

stratigos:
The test environment is unable to read an `.ex` file
located in the `config/` dir by default for Phoenix
apps as of v1.7
    
This change makes a small edit to the README to suggest
the `sandbox.ex` file be placed in the `test/support` dir.


stratigos:
Im going to close this and reopen with a branch that lacks merge commits, I sync'd with upstream poorly.

PR #766, opened 2024-04-04
Update README: Local Sandbox File Location

stratigos:
The test environment is unable to read an `.ex` file located in the `config/` dir by default for Phoenix apps as of v1.7

This change makes a small edit to the README to suggest the `sandbox.ex` file be placed in the `test/support` dir.


stratigos:
I cannot imagine that CI is breaking due to a README change. Is there a way I can re-run the checks? Are these flaky test issues that will resolve after a few retries?

PR #770, opened 2024-05-07
Make Query.text/2 docs also point to assert_text/{2,3}

s3cur3:
This beefs up the docs on `Query.text/2`, which a lot of examples combine with `assert_has/2` for asserting that a particular element has some text. Pointing to `assert_text/{2,3}` from the docs here would have resolved a recent issue Mitch kindly helped me with [in Slack](https://elixir-lang.slack.com/archives/C4H1XRC0J/p1715093283885629) where I was trying to assert on the text of multiple nested elements.

### DIFF
--- a/lib/wallaby/query.ex
+++ b/lib/wallaby/query.ex
@@ -207,7 +207,7 @@ defmodule Wallaby.Query do
   Test:
 
       ```
-      assert_has_text(session, "Unread messages: 1 message")
+      assert_text(session, "Unread messages: 1 message")
 
       notifications_block = Query.css("#unread-notifications")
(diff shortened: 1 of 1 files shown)


lib/wallaby/query.ex
```diff
@@ -170,14 +170,49 @@ defmodule Wallaby.Query do
...
+      <div id="unread-notifications">
+        Unread messages:
+        <strong>1 message</strong>
+      </div>
+      ```
+
+  Test:
+
+      ```
+      assert_has_text(session, "Unread messages: 1 message")
```
mhanberg:
```suggestion
      assert_text(session, "Unread messages: 1 message")
```

PR #774, opened 2024-05-17
Trapping process causes conflict with created users

savhappy:
When attempting to create a user to login during Wallaby test, this error appears:

`Wallaby.SessionStore.handle_info({:EXIT, #PID<0.984.0>, :normal}, %{ets_table: :session_store})`

Switching Process.flag(:trap_process, true) to Process.flag(:trap_process, false) solves problem.

### DIFF
--- a/lib/wallaby/session_store.ex
+++ b/lib/wallaby/session_store.ex
@@ -33,7 +33,7 @@ defmodule Wallaby.SessionStore do
       if(name == :session_store, do: [:named_table], else: []) ++
         [:set, :public, read_concurrency: true]
 
-    Process.flag(:trap_exit, true)
+    # Process.flag(:trap_exit, true)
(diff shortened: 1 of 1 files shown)


lib/wallaby/session_store.ex
```diff
@@ -33,7 +33,7 @@ defmodule Wallaby.SessionStore do
       if(name == :session_store, do: [:named_table], else: []) ++
         [:set, :public, read_concurrency: true]
 
-    Process.flag(:trap_exit, true)
+    Process.flag(:trap_exit, false)
```
mhanberg, line 36: You can just delete the line

mhanberg (changes requested):

mhanberg:
There are some unrelated failures that might be from chrome driver changes. 

I'll take a look

PR #776, opened 2024-06-06
Address deprecation; prefer ExUnit.Case.register_test/6

vanderhoop:
- `ExUnit.Case.register_test/6` has been available since Elixir 1.11.0, and now triggers deprecation warnings as of Elixir 1.17. This would be a breaking change if the intention is to support Elixir versions prior to 1.11, but I think we're all good on that front, given the existing CI version matrix


mhanberg:
Thanks!

PR #788, opened 2024-08-22
feat: allow to slow down browser actions

bagrat:
For cases mentioned in https://github.com/elixir-wallaby/wallaby/issues/337 and in order to enable the developer to watch their test execute in a comprehensible way while developing, I've added the `:slow_down` config option to implement that feature.

I'm not sure how to test this, but open for suggestions and recommendations.

### DIFF
--- a/lib/wallaby/chrome.ex
+++ b/lib/wallaby/chrome.ex
@@ -397,6 +397,12 @@ defmodule Wallaby.Chrome do
 
   defp delegate(fun, element_or_session, args \\ []) do
     check_logs!(element_or_session, fn ->
+      delay = Application.get_env(:wallaby, :slow_down, 0)
+
+      if delay > 0 do
+        :timer.sleep(delay)
+      end
+
       apply(WebdriverClient, fun, [element_or_session | args])
     end)
   end


mhanberg:
Hi! I have a couple questions/comments

Q: Are there any similar features in similar libraries for other languages?

C: this is just implemented in the chrome driver adapter, it would need to be useable for all adapters

C: features like this should be implemented in a way that you can pass the option to start_session, and fall back to app config, then fall back to a default. And in this case, potentially even pass it to individual functions

bagrat:
Hey, thanks for a quick response!

> Q: Are there any similar features in similar libraries for other languages?

Yes, I've used them both in Puppeteer and Playwright before, it's called [`slowMo`](https://playwright.dev/docs/debug#headed-mode).

> C: this is just implemented in the chrome driver adapter, it would need to be useable for all adapters

Could you possible point me in the right direction where this delay should also be put in?

> C: features like this should be implemented in a way that you can pass the option to start_session, and fall back to app config, then fall back to a default. And in this case, potentially even pass it to individual functions

[In chrome.ex](https://github.com/bagrat/wallaby/blob/slow-down/lib/wallaby/selenium.ex#L102) and [selenium.ex](https://github.com/bagrat/wallaby/blob/slow-down/lib/wallaby/selenium.ex#L102)?

PR #792, opened 2024-12-12
Make assert_text/3 pipable

bartblast:
Currently, assert_text/3 returns a %Wallaby.Element{} struct matched inside the find(parent, query) call, instead of the original parent passed to assert_text/3 as specified in the function's doc and spec.


mhanberg:
Test failures seem unrelated. I'll take a look before merging.

PR #795, opened 2024-12-29
Only spawn a browser when using the `feature` macro, not normal `test`s

### LINKED ISSUE #794: `use Wallaby.Feature` spawns an instance of Chromedriver for every `test` in the module, not just `feature`s
Hi @mhanberg! I'm happy to submit a patch for this if you'd like. It's probably a breaking change for some people who were accidentally using `test`, but based on my reading of the docs, one shouldn't expect a `:session` to exist without the `feature` macro.

### Elixir and Erlang/OTP versions

Erlang/OTP 27 [erts-15.1.1] [source] [64-bit] [smp:16:16] [ds:16:16:10] [async-threads:1] [jit] [dtrace]

Elixir 1.17.3 (compiled with Erlang/OTP 27)

### Operating system

macOS Sequoia 15.2

### Browser

Chrome

### Driver

Chromedriver

### Correct Configuration

- [X] I confirm that I have Wallaby configured correctly.

### Current behavior

This set of 100 tests takes about 0.01 seconds to run:

```elixir
defmodule WallabyReproCase do
  use ExUnit.Case, async: true

  for idx <- 1..100 do
    test "this is fast (#{idx})" do
      assert true
    end
  end
end
```

But this set of 100 tests takes 26.9 seconds, and as it's running, I see Chrome launching repeatedly in Activity Monitor.:

```elixir
defmodule WallabyReproCase do
  use ExUnit.Case, async: true
  use Wallaby.Feature

  for idx <- 1..100 do
    test "this is sloooooow (#{idx})" do
      assert true
    end
  end
end
```

Of course, this is expected based on the `__using__` definition in `Wallaby.Feature`, because the Chromedriver session is being launched in the `setup` hook, which will apply to all the tests in the module.

### Expected behavior

The latter set of tests should be more or less as fast as the former, and Chrome shouldn't be launched at all.

My suggestion would be that the `setup` hook in `Wallaby.Feature` check the test context for the `:feature` tag and not launch the browser driver unless it is `true`.

### Test Code & HTML

(See above)

### Demonstration Project

_No response_

s3cur3:
This fixes an unintended behavior where having `use Wallaby.Feature` at the top of your module would cause a browser to be spawned for every test in the module, not just `feature` tests. The result is that if you had a large test module, most of which used the normal `test` macro and one of which used the `feature` macro, every test would take a minimum of a third of a second or more.

~I also took the liberty to check and see in the `setup` block if we already had a `:session`, and if so, avoid spawning a second. I don't *think* anyone was relying on the behavior where a second browser was being spawned if you had `use Wallaby.Feature` at both the top of the module and a `describe` block.~

Resolves #794

### DIFF
--- a/lib/wallaby/feature.ex
+++ b/lib/wallaby/feature.ex
@@ -23,14 +23,7 @@ defmodule Wallaby.Feature do
       import Wallaby.Feature
 
       setup context do
-        # Only set up a session if this is a feature test (using the `feature` macro).
-        feature_test? = context[:feature] == true
-
-        # If you have `use Wallaby.Feature` at the top of your test module,
-        # then also within a `describe` block, we only want to set up one session.
-        already_has_session? = is_map_key(context, :session)
-
-        if feature_test? and not already_has_session? do
+        if context[:test_type] == :feature do
           metadata = unquote(__MODULE__).Utils.maybe_checkout_repos(context[:async])
 
           start_session_opts =
--- a/lib/wallaby/feature.ex
+++ b/lib/wallaby/feature.ex
@@ -23,22 +23,33 @@ defmodule Wallaby.Feature do
       import Wallaby.Feature
 
       setup context do
-        metadata = unquote(__MODULE__).Utils.maybe_checkout_repos(context[:async])
+        # Only set up a session if this is a feature test (using the `feature` macro).
+        feature_test? = context[:feature] == true
 
-        start_session_opts =
-          [metadata: metadata]
-          |> unquote(__MODULE__).Utils.put_create_session_fn(context[:create_session_fn])
+        # If you have `use Wallaby.Feature` at the top of your test module,
+        # then also within a `describe` block, we only want to set up one session.
+        already_has_session? = is_map_key(context, :session)
 
-        get_in(context, [:registered, :sessions])
-        |> unquote(__MODULE__).Utils.sessions_iterable()
-        |> Enum.map(fn
-          opts when is_list(opts) ->
-            unquote(__MODULE__).Utils.start_session(opts, start_session_opts)
+        if feature_test? and not already_has_session? do
+          metadata = unquote(__MODULE__).Utils.maybe_checkout_repos(context[:async])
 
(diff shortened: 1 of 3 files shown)


lib/wallaby/feature.ex
```diff
@@ -23,22 +23,33 @@ defmodule Wallaby.Feature do
...
-          i when is_number(i) ->
-            unquote(__MODULE__).Utils.start_session([], start_session_opts)
-        end)
-        |> unquote(__MODULE__).Utils.build_setup_return()
+        # Only set up a session if this is a feature test (using the `feature` macro).
+        feature_test? = context[:feature] == true
+
+        # If you have `use Wallaby.Feature` at the top of your test module,
+        # then also within a `describe` block, we only want to set up one session.
+        already_has_session? = is_map_key(context, :session)
```
mhanberg, line 31: Let's not deal with this in this pull request and do a second one. 

This might be better to emit a warning rather than be silent.

lib/wallaby/feature.ex
```diff
@@ -23,22 +23,33 @@ defmodule Wallaby.Feature do
...
-        |> Enum.map(fn
-          opts when is_list(opts) ->
-            unquote(__MODULE__).Utils.start_session(opts, start_session_opts)
-
-          i when is_number(i) ->
-            unquote(__MODULE__).Utils.start_session([], start_session_opts)
-        end)
-        |> unquote(__MODULE__).Utils.build_setup_return()
+        # Only set up a session if this is a feature test (using the `feature` macro).
+        feature_test? = context[:feature] == true
```
mhanberg, line 27: I think here you want to check `context[:test_type] == :feature`

mhanberg: Also I would just inline this. It's self explanatory.

mhanberg (changes requested):

mhanberg:
The tests might fail a bunch but you don't have to worry about it.

s3cur3:
Changes made! ☺️

mhanberg (approved):

PR #810, opened 2025-05-02
Feature: Allow flag for disabling sandbox

Zurga:
Sometimes the Ecto.Adapters.SQL.Sandbox is not a good fit when testing. If Postgres triggers are setup, the sandbox will not trigger them because the transaction is not comitted. For that reason it might be good to be able to disable the sandbox on a Case basis. 

This PR adds a `:sandbox` opt to `use Wallaby.Feature` allowing the caller to disable the sandbox, even if they have the module loaded.

### DIFF
--- a/lib/wallaby/feature.ex
+++ b/lib/wallaby/feature.ex
@@ -13,9 +13,13 @@ defmodule Wallaby.Feature do
   ```
   config :wallaby, otp_app: :your_app
   ```
+  When used, it accepts the following options:
+   * `:sandbox` - whether or not to use the sandbox in the tests. Defaults to `true`.
   """
 
-  defmacro __using__(_) do
+  defmacro __using__(opts \\ []) do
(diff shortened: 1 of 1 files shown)


mhanberg:
If you don't want to automatically checkout the ecto repos, you might be better off just importing Wallaby.Feature instead of "use" it. 

That would require starting the session manually, however

Zurga:
Yes, I saw this in the tests but then I also lose out on the automatic
screenshot on failure, capabilities and other nice stuff `use
Wallaby.Feature` provides

Op vr 2 mei 2025 2:42 p.m. schreef Mitchell Hanberg <
***@***.***>:

mhanberg:
You still get automatic screenshots if you use the `feature` macro, which you get by importing Wallaby.Feature

Zurga:
Ah yes you are correct! If you are not inclined to merge this PR, feel free to close it

PR #834, opened 2026-01-11
fix: bug in shutdown — pid2 is assigned from  $1 instead of $2

101v:
pid2 is assigned from  $1 instead of $2 , therefore pid2 isn’t used correctly.


mhanberg:
Can you reproduce the problem this causes?

I noticed this years ago and changed it but seemed to notice it cause it to not work correctly. 

I'm interested to see what you've observed.

